PR #11, opened 2026-08-27
Add noscript banner for browsers with JavaScript disabled

2140data:
If you open the page with JavaScript turned off you just get the header, the warnings and the three tab buttons with no forms. The panels are hidden in the HTML and only shown by `app.js`, so with JS off there's nothing on the page telling you why it's empty.

This adds a `<noscript>` block above the beta warning in `src/index.html`.

Rebuilt `index.html` and `entropylab-0.1.3.html` with `npm run build`.

Open `index.html` with JS disabled and the banner shows at the top. With JS enabled, it doesn't appear.


itme-brain:
ACK

PR #18, opened 2026-08-27
Tighten horizontal layout and align form fields

2140data:
The choice cards were stacked full-width even on desktop, which made those sections tall and hard to read.

This displays them in a responsive grid. Phone/tablet still stacks vertically.

Also fixed the PSBT session-key textarea so it lines up with the field next to it.

`npm run ci` passes, generated files rebuilt.

screenshot


portlandhodl:
tAck

PR #30, opened 2026-08-28
Recognize Bitcoin Core-style low-r grinding in PSBT nonce checks

MrHodlX:
Bitcoin Core grinds RFC 6979 extra entropy until r is low (first byte < 0x80), saving one DER byte. EntropyLab currently treats those honest Core signatures as a plain RFC 6979 mismatch.

This teaches the PSBT nonce check to:
- accept plain RFC 6979
- accept Core-style low-r grind retries (little-endian 32-bit counter as extraEntropy)
- accept 32 zero extra-entropy bytes used by some libraries
- still warn on anything else, without calling it a compromise

A mismatch alone is not evidence of a compromised signer. Reused r on two different messages is the real alarm. Session key stays in RAM only.

npm run test:lowr covers the grind counter, low-r byte test, and copy.


portlandhodl:
tAck

PR #31, opened 2026-08-28
Add Jade anti-exfil (sign-to-contract) transcript checks to PSBT inspect

MrHodlX:
Offline host-side verify of secp256k1-zkp anti_exfil_host_verify: paste the USB host nonce ρ and signer opening R next to a PSBT. No private key required. Does not speak USB/QR to Jade and does not implement BitBox anti-klepto.


portlandhodl:
Needs fix

`
⚠️ Bug: malformed opening crashes the entire PSBT report
hodlParseAntiExfil only checks the opening's prefix byte (02/03). A 33-byte value with a valid prefix but an x-coordinate not on the curve passes parsing, then Point.fromBytes → assertValidity() throws inside hodlAntiExfilCommitOk, which has no try/catch. The exception propagates through hodlRenderPsbt, so one bad transcript wipes out the whole inspection report (empty output, cryptic noble error) instead of flagging just that signature. Recommended fix: wrap the hodlAntiExfilCommitOk call in try/catch and degrade to a psbt-warn row, like the existing RFC 6979 branch does.
`

PR #33, opened 2026-08-28
Add Taproot BIP86 multisig wallets

ottoz0r:
Adds Taproot as a multisig script type, next to Legacy / Nested SegWit / Native SegWit.

Key derivation on Taproot · BIP86 now emits a paste-ready co-signer export:

`[fingerprint/86h/coinh/accounth]xpub…`

Multisig accepts those origins, and also BIP48 script-path keys at `48h/…/3h`. Script type auto-selects when the co-signer keys agree.

The watch-only descriptor is script-path only:

`tr(<BIP341 NUMS>,sortedmulti_a(m,keys/<0;1>/*))#checksum`

Internal key is the BIP341 NUMS point, so spending is only possible through the `multi_a` leaf. Addresses are `bc1p…`. The existing descriptor QR is used as-is.

`test/descriptor.test.mjs` and `test/ui-defaults.test.mjs` cover origin detection, the NUMS/`sortedmulti_a` payload, and the new script-type option. A local 2-of-2 from the abandon and zoo 12-word vectors derived `bc1psm7mx7ng7wc52eetrsnk0z864hdqpjqtfrw0vpvs2jyal59s5g7ql267q4`. Sparrow/signer import is not in this PR.

### DIFF
--- a/entropylab-0.1.3.html
+++ b/entropylab-0.1.3.html
@@ -816,8 +816,33 @@ input.bad, textarea.bad { border-color: var(--danger) !important; }
   transform: translateX(-50%); background: var(--border);
 }
 .msig-threshold-help { margin: 12px 0 0; text-align: center; }
-.msig-keys { display: grid; gap: var(--space-component); margin-top: var(--space-section); }
-.msig-keys .field { margin-top: 0; }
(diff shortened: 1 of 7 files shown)


portlandhodl:
BIP 86 isn't showing errors when using a non-bip 86 compatible derivation path.

PR #35, opened 2026-08-28
Add extension interface with documented EntropyLab API

portlandhodl:
Adds a documented way for third-party extensions to plug into EntropyLab:

- **`EntropyLab` global API** — `registerExtension()` with a validated manifest (`id`, `name`, `version`, `author`, `description`, `activate`), plus `listExtensions()`, `unregisterExtension()`, `validateManifest()`, and `loadSource()`.
- **Extensions workspace** — fourth tab; loads a local `.js` file or pasted source, shows the SHA-256 of each extension for checksum verification, and lists status/metadata with Remove.
- **First hook type** — `api.addPanel({ title, render })` renders custom panels inside the Extensions workspace.
- **Security model** — documented in `docs/EXTENSIONS.md`: no sandbox; extensions must be audited and trusted like the app itself, and the loader encourages comparing the displayed SHA-256 against the author's published checksum.
- Load path uses inline `<script>` injection (CSP-safe under `script-src 'unsafe-inline'` without `unsafe-eval`).

Tests: `test/extensions.test.mjs` (8 format/registry tests) wired into `test:ci`; `npm run ci` is green (148 tests + verify).


itme-brain:
Approach NACK

PR #36, opened 2026-08-28
feat: implement dice fairness test

dani69654:
## Summary
- Adds a live die-fairness panel to Dice rolls that runs Pearson’s χ² goodness-of-fit on the entered faces while you type.
- Uses the die implied by the selected method (D6 for hashed rolls, D4 + coin for BitBox, D8 + D16 for D++) instead of guessing the die type. Verdicts follow the [Dice Fairness Tester](https://dicefairness.johnellmore.com/) thresholds (CDF < 80% fair, < 90% unsure, ≥ 90% biased) and the five-rolls-per-face rule from [RPG Stack Exchange](https://rpg.stackexchange.com/questions/70802/how-can-i-test-whether-a-die-is-fair).
- The check is advisory only: a biased-looking die does not block Derive Wallet. χ² CDF is computed locally (no jstat), so the air-gapped HTML stays self-contained.

image


portlandhodl:
Needs to be by default hidden, and user can use a button (glyph = text) to expand.

dani69654:
@portlandhodl implemented the change and rebased

PR #63, opened 2026-08-29
BitBox and D++ dice: coin-side labels, keypad layout, copy fixes

dplusplus1024:
Follow-up to #52, covering the dice keypads and their copy.

### DIFF
--- a/src/css/styles.css
+++ b/src/css/styles.css
@@ -240,11 +240,13 @@ label.field { display: block; margin-top: var(--space-component); font-size: 14p
   .choice-grid.three-up { grid-template-columns: repeat(3, minmax(0, 1fr)); }
 }
 .dice-input-pad { display: grid; grid-template-columns: repeat(3, minmax(0, 1fr)); gap: 8px; margin-top: var(--space-control); }
-.dice-input-pad.with-coin { grid-template-columns: repeat(3, minmax(0, 1fr)) minmax(0, 3fr); }
+.dice-input-pad.faces-1-6 { grid-template-columns: repeat(3, minmax(0, 1fr)); }
+@media (min-width: 720px) {
+  .dice-input-pad.faces-1-6 { grid-template-columns: repeat(6, minmax(0, 1fr)); }
+}
 .dice-input-pad.dplus {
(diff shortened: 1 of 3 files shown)


portlandhodl:
`
 Critical: typed h/t coin flips silently derive the wrong wallet
The PR relabels the BitBox convention to 1–3 = Tails, 4–6 = Heads (keypad captions in hodlUpdateDiceButtons, all help copy) and updates the numeric path to match:
let coin = input === "t" || input === "1" || input === "2" || input === "3" ? 0 : 1;
But the typed-letter path still uses the old convention: h → bit 0, t → bit 1. Verified behaviorally against the PR head:
faces 1,2,3 → bit 0   faces 4,5,6 → bit 1      (new, consistent)
typed "h"   → bit 0   typed "t"   → bit 1      (old, inconsistent)
So a user who reads the new UI ("1–3 is Tails") and types t for a low face gets the opposite bit — a different seed phrase with no warning. Two users entering the same physical rolls via keypad vs. keyboard derive different wallets. That's a fund-loss-at-recovery class bug: the wallet works, but writing down "tails" semantics and re-entering later reproduces the wrong seed.
Fix: swap to input === "h" || input === "1" ... ? 0 : 1... wait — no: bit 0 must mean Tails per the new labels, so it should be input === "t" || low faces ? 0 : 1 for the digit part, but h must map to 1 and t to 0 — i.e. the correct full fix is input === "h" ? 1 : input === "t" ? 0 : (face <= 3 ? 0 : 1).
`

dplusplus1024:
Thanks for digging into this — but the mapping is already consistent on the PR
head: `t` and faces 1-3 both give bit 0, `h` and 4-6 both give bit 1. The
`h` -> `t` swap was part of this PR. Your suggested fix produces the same truth
table as what's committed.

The typed path was unreachable anyway — `hodlSanitizeDiceInput` keeps only
characters 1-6, so typing `h`/`t` yields no word at all.

You were right that it was a hazard to read, though, so I've removed the dead
letter handling; the bit now comes from the digits alone. Behaviour verified
unchanged, suite green.

PR #74, opened 2026-08-29
Show a LifeHash icon beside each master fingerprint

portlandhodl:
## What

Every master fingerprint the app shows — the **Base seed** and **With passphrase** preview cards — now sits next to its deterministic [LifeHash](https://lifehash.info) icon, so two keys can be told apart at a glance instead of by reading eight hex digits.

## How

New `src/js/lifehash.js`: a faithful, **dependency-free** port of the LifeHash (version2) reference algorithm:

- SHA-256 (WebCrypto, no vendored lib) seeds a 16×16 Conway's Game of Life on a toroidal grid.
- The run's history collapses into a normalised grayscale "frac" grid.
- Hash bits deterministically pick the colour gradient (monochromatic / complementary / triadic / analogous over the CMYK-safe spectrum) and the snowflake/pinwheel symmetry.
- Rendered to a crisp nearest-neighbour PNG as a `data:` URL — works fully offline inside the existing CSP (`img-src 'self' data:`), and per the LifeHash presentation guidance (`image-rendering: pixelated`, square, no vignette).

The fingerprint preview cards gain an `` next to the hex; `hodlSetMasterFingerprintCard` resolves the icon asynchronously and only sets `src` once the value is current (stale async results are discarded).

## Correctness — pinned to the reference

`test/lifehash.test.mjs` reproduces the **exact 32×32 RGB output** of the canonical `lifehash` JS package (AndreasGassmann/lifehash, itself a port of the Blockchain Commons reference) for four fingerprints — the port is bit-for-bit identical, not just visually similar. Also covered: PNG structure (96×96 at module size 3, 8-bit truecolour), determinism, lowercase normalisation, distinctness, and digest-length validation.

The build inlines the module behind a new `@@JS_LIFEHASH@@` token; `test:lifehash` is added and wired into `test:ci`.

## Verified

- `npm run test:ci` — **216/216 pass** (208 existing + 6 LifeHash + 2 new UI assertions).
- `npm run verify` — OK.
- Rendered PNG decoded and visually confirmed: symmetric Game-of-Life structure with a smooth gradient.
- No new dependencies; `package.json` gains only the test script.


HodlDee:
@portlandhodl — a provenance question about this PR.

In PR #74, `src/js/lifehash.js` is described as a "faithful port of the reference algorithm (Blockchain Commons / the lifehash JS package)." Which source materials were consulted when writing it? Was any Blockchain Commons C++ or Andreas Gassmann TypeScript source translated or adapted, or was the implementation written from the algorithm description and validated against test vectors?

We maintain a downstream adaptation and need to determine whether the BSD-2-Clause-Patent or MIT notices must accompany it, or whether EntropyLab's public-domain contribution terms cover the complete implementation.

portlandhodl:
@DeesNeez Precise answer, after a provenance review of the code:

**Sources consulted.** The reference implementations' source code: primarily the [AndreasGassmann/lifehash](https://github.com/AndreasGassmann/lifehash) TypeScript/JavaScript package (MIT), which is itself a port of [BlockchainCommons/bc-lifehash](https://github.com/BlockchainCommons/bc-lifehash) (C++, BSD-2-Clause-Patent). Only the `version2` subset was implemented.

**Adapted, not clean-room.** `src/js/lifehash.js` was written by following those sources function-by-function and re-expressing them in this project's idiom (closures instead of classes, flat arrays instead of grid objects, WebCrypto SHA-256, a hand-rolled PNG encoder) — it was not written from the prose algorithm description alone. The evidence is in the code: the verbatim error string `"BitEnumerator underflow."`, function names taken from the TS package (`runGameOfLife`, `buildFracGrid` — the C++ inlines that logic), and TS-port-specific idioms (e.g. `sha256(data).toString()` as the history-set key). Validation came after: the pinned vectors reproduce the canonical package's exact 32×32 RGB output, and `fuzzing/lifehash/` differentially compares against the npm `lifehash` package directly.

**What that means for your downstream.** Carry both notices: MIT (Copyright (c) 2022 Andreas Gassmann) and BSD-2-Clause-Patent (Copyright (c) 2019 Blockchain Commons, LLC). EntropyLab's public-domain terms cover the project's original code; they cannot cover adapted upstream expression, so they do not cover the complete `lifehash.js` implementation. Both licenses are permissive — retaining two short notices is the entire obligation, and the BSD-2-Clause-Patent notice brings a patent grant with it. (One nuance for your counsel: a strict abstraction–filtration analysis might conclude the surviving literal expression is de minimis and that no notice is strictly required — but given the file describes itself as a faithful port of these sources, retaining the notices is the defensible course.)

#265 adds both notices in full to the `lifehash.js` header — which the build inlines verbatim into `entropylab.html`, so the artifact carries them too — and documents the exception in CONTRIBUTING.md. Downstream redistributors of the file or the artifact should keep those notices intact.

HodlDee:
Thank you for the detailed provenance review. Your answer resolves our question. We'll carry both the MIT notice for Andreas Gassmann's implementation and the BSD-2-Clause-Patent notice for Blockchain Commons in our LifeHash source and generated artifacts. Please disregard my previous wording that implied the Blockchain Commons notice alone was sufficient. No further response is needed.

PR #90, opened 2026-08-29
Fix decimal D16 transcript tokenization

### LINKED ISSUE #86: Decimal D16 input is tokenized as hexadecimal characters
## Summary

D++ offers a `Decimal D16 - 1-16` mode and displays examples containing multi-character outcomes such as `1 16 16`. The parser does not tokenize decimal results. It removes separators and treats each remaining character as one roll, then decodes each D16 character as hexadecimal.

As a result, decimal outcomes 10 through 16 become two separate rolls. Every subsequent roll position shifts, and the application can derive a different but checksum-valid mnemonic from the user's physical transcript. It can also consider a seed complete after fewer physical rolls than the documented workflow requires.

## Affected Code

- `src/js/app.js:967-970` accepts exactly one hexadecimal character as a D16 value.
- `src/js/app.js:981-989` skips separators and emits one token per character.
- `src/js/app.js:991-993` leaves the transcript unchanged when switching notation.
- `src/js/app.js:995-1046` groups every three characters as D8, D16, D16 and determines completion from character counts.
- `src/js/app.js:1723-1728` advertises decimal values 1 through 16 and shows multi-character decimal examples.

## Security Impact

An honest user can fund a wallet whose mnemonic cannot later be reconstructed by applying the displayed decimal D16 convention to the recorded physical rolls. If the separately generated mnemonic was not preserved, recovery follows the documented transcript to a different wallet and the original funds may be permanently inaccessible.

Because multi-character values inflate the parser's roll count, the derived wallet can also contain materially less physical entropy than the selected phrase size implies.

## Preconditions

- The user selects D++ and Decimal D16.
- The user types decimal outcomes, including at least one value from 10 through 16, rather than using the on-screen keypad's internal hexadecimal values.
- The user trusts the valid word preview and later relies on the physical roll transcript for recovery.

## Synthetic Reproduction

1. Select a 12-word D++ seed and Decimal D16.
2. Enter the physical group D8=`1`, D16=`16`, D16=`16` as displayed: `1 16 16`.
3. Under the displayed convention, each `16` represents underlying D16 value zero, so the group should select BIP39 index 0 (`abandon`).
4. The current tokenizer instead reads the first three characters as `1`, `1`, `6`, selecting index `(1-1)*256 + 1*16 + 6 = 22` (`actress`). The remaining characters begin the next group.
5. Compare complete decimal and hexadecimal transcripts representing the same synthetic physical rolls and observe that they do not select the same mnemonic.

## Recommended Fix

Use a delimiter-aware tokenizer in decimal mode. Treat each integer from 1 through 16 as exactly one roll, convert 16 to canonical hexadecimal `0`, and convert 10 through 15 to `A` through `F`. Reject ambiguous or out-of-range decimal input.

Alternatively, remove decimal text entry and clearly state that only hexadecimal transcript values are accepted. Switching conventions should translate complete tokens or reject ambiguous existing input rather than silently retaining it.

## Acceptance Criteria

- Decimal `1 16 16` selects the same word as hexadecimal `100`.
- Values 10 through 16 each increment the physical roll count by exactly one.
- Incomplete physical transcripts cannot enable derivation.
- Complete 12-, 18-, and 24-word decimal and hexadecimal vectors representing the same rolls select identical mnemonics.
- Invalid values such as decimal `0`, decimal `17`, and hexadecimal letters in decimal mode are rejected unambiguously.

Jdelg718:
Fixes #86.

## Summary

- tokenize decimal D16 results as delimiter-separated whole rolls from 1 through 16
- keep decimal and hexadecimal transcripts equivalent across every supported BIP39 phrase length
- reject decimal 0, 17, hexadecimal letters, ambiguous notation changes, incomplete transcripts, and excess rolls
- make the decimal keypad emit and validate actual decimal tokens
- preserve delimiters and caret positions during incremental entry and notation conversion
- include the D++ regression suite in `test:ci`

## Security rationale

A physical D16 result from 10 through 16 must count as exactly one roll. Reading those values character-by-character shifts every later roll and can derive a different but checksum-valid wallet from the recorded transcript. The parser now uses the selected notation explicitly and refuses ambiguous or out-of-range input rather than silently changing the represented entropy.

## Verification

- `node --test test/dplus.test.mjs` — 14/14 passed
- `npm run ci` — 262/262 tests passed; build and site verification passed
- independent pre-commit review — passed with no security concerns or logic errors

The full `npm test` suite was also exercised locally; its only failure was the existing headless-browser requirement because Firefox is not installed on the test machine.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
     "build": "node scripts/build.mjs",
     "clean": "node scripts/build.mjs --clean",
     "test": "node --test test/*.test.mjs",
-    "test:ci": "node --test test/network-check.test.mjs test/browser-check.test.mjs test/ui-defaults.test.mjs test/validate.test.mjs test/descriptor.test.mjs test/seedqr.test.mjs test/cards.test.mjs test/copy-seed.test.mjs test/number-bases.test.mjs test/seed-numbers.test.mjs test/psbt-low-r.test.mjs test/psbt-metadata.test.mjs test/secret-clear.test.mjs test/sqlite-writer.test.mjs test/wallet-export.test.mjs test/psbt-anti-exfil.test.mjs test/address-match.test.mjs test/psbt-nonce-reuse.test.mjs test/lifehash.test.mjs",
+    "test:ci": "node --test test/network-check.test.mjs test/browser-check.test.mjs test/ui-defaults.test.mjs test/validate.test.mjs test/descriptor.test.mjs test/seedqr.test.mjs test/cards.test.mjs test/copy-seed.test.mjs test/number-bases.test.mjs test/seed-numbers.test.mjs test/dplus.test.mjs test/psbt-low-r.test.mjs test/psbt-metadata.test.mjs test/secret-clear.test.mjs test/sqlite-writer.test.mjs test/wallet-export.test.mjs test/psbt-anti-exfil.test.mjs test/address-match.test.mjs test/psbt-nonce-reuse.test.mjs test/lifehash.test.mjs",
     "test:network": "node --test test/network-check.test.mjs",
     "test:browser-check": "node --test test/browser-check.test.mjs",
     "test:ui": "node --test test/ui-defaults.test.mjs",
--- a/src/js/app.js
+++ b/src/js/app.js
@@ -1847,9 +1847,9 @@ function hodlDPlusStepBits(step) {
 function hodlDPlusStepLabel(step) {
   return step === "d8" ? "D8" : step === "d16" ? "D16" : "a coin flip";
 }
-function hodlDPlusStepValue(step, face) {
+function hodlDPlusStepValue(step, face, numberedD16 = false) {
   if (step === "d8") return /^[1-8]$/.test(face) ? Number(face) - 1 : null;
-  if (step === "d16") return hodlDPlusD16Value(face);
+  if (step === "d16") return hodlDPlusD16Value(face, numberedD16);
   return /^[1-8]$/.test(face) ? (Number(face) >= 5 ? 1 : 0) : null;
 }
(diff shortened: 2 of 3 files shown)


w-s-bitcoin:
Thank you for identifying the decimal tokenizer problem and for adding focused regression coverage. After comparing the implementation with the D++ / Keysa workflow, we should not merge this version because its decimal-to-hex mapping is incompatible with the published method.

The workflow uses one D8 followed by two hexadecimal D16 faces. A numbered D16 labeled 1 through 16 must be converted by subtracting one: 1 maps to 0, 10 maps to 9, 11 maps to A, and 16 maps to F. Therefore physical rolls 1,1,1 correspond to canonical transcript 100 and BIP39 word abandon; 8,16,16 correspond to 8FF and zoo. This PR instead maps 1 to 1 and 16 to 0, which rotates every D16 value and derives different seed words.

Sources:
- https://thesimplestbitcoinbook.net/wp-content/uploads/2023/09/Roll-Your-Own-Seed-Phrase-PDF.pdf
- https://github.com/reardencode/bip39_dice/blob/master/README.md

I am closing this in favor of a smaller replacement that removes the decimal/hex mode ambiguity, accepts the canonical hex transcript directly, and retains regression tests for the verified D++ vectors.

Jdelg718:
Thanks for the careful review and source check. You’re right—the mapping in our patch followed the issue’s stated `16 → 0` convention, but that does not match the published D++/Keysa method.

Removing decimal mode and accepting one canonical `0–F` transcript is the safer and simpler solution, especially where notation ambiguity can derive an entirely different wallet. I’m glad the verified vectors and D++ CI coverage carried forward into #97.

Appreciate the clear explanation and the thoughtful resolution. We learned something useful here—thanks again.

PR #91, opened 2026-08-29
Fix #86: tokenise D++ decimal D16 by delimiter, not by character

### LINKED ISSUE #86: Decimal D16 input is tokenized as hexadecimal characters

tmmmemcee:
## Summary

Fixes #86. The D++ tokenizer split every non-separator character into a single token, which meant decimal values 10-16 (multi-character) were broken into separate rolls. `"1 16 16"` became `[1,1,6,1,6]` and selected "actress" (BIP39 index 22) instead of "abandon" (index 0). An honest user could fund a wallet whose mnemonic could not be reconstructed from the recorded physical rolls.

## What changes

- `hodlDPlusTokens` now takes a `numberedD16` parameter and, when set, splits on whitespace and the existing separator set, then normalises each chunk: 1-9 → themselves, 10-15 → A-F, 16 → 0.
- Out-of-range or non-numeric chunks keep their face so the UI can show what the user typed, plus an `invalid: true` flag so the existing pipeline surfaces them via `invalidRanges`.
- Decimal mode requires D++'s documented decimal notation. Hex transcripts continue to tokenise exactly as before, character-by-character.

## Tests

Adds `test/dplus-decimal.test.mjs` with 5 tests covering the issue's synthetic repro and all four acceptance criteria (decimal/hex equivalence for every target size, single-roll accounting for values 10-16, unambiguous rejection of out-of-range and hex-letter input in decimal mode).

```
npm run test:ci   # 248/248 pass
npm run build     # deterministic from src/
```

The single failure in `npm test` is the headless-Firefox suite timing out in this environment (`RenderCompositorSWGL failed mapping default framebuffer`); it was failing before this change and is unrelated to the parser.

## Intentionally excluded

- `entropylab.html` (build artifact; CI rebuilds and commits it after merge per the README workflow — keeping it out of PRs avoids merge conflicts and matches the repo's existing convention).
- `hodlSanitizeDPlusInput` — needs `numberedD16` threaded through so the sanitiser respects decimal mode. Split into a follow-up so this PR stays focused on the parser, per CONTRIBUTING.md's "small and focused, one change each".

### DIFF
--- a/src/js/app.js
+++ b/src/js/app.js
@@ -1881,10 +1881,30 @@ function hodlDPlusD16Value(face) {
 }
 // Single tokenizer shared by the parser and the input sanitiser so the two can
 // never disagree about where one roll ends and the next begins.
-function hodlDPlusTokens(value) {
+function hodlDPlusTokens(value, numberedD16 = false) {
   let text = String(value ?? ""),
     entries = [],
     index = 0;
+  if (numberedD16) {
+    // Decimal D16: each non-separator chunk is one roll. Normalise 1-9 to
+    // themselves, 10-15 to A-F, 16 to 0. Out-of-range or non-numeric chunks
+    // keep their original face but are flagged invalid so the downstream
+    // pipeline surfaces them via invalidRanges.
+    for (const match of text.matchAll(/[^\s,;|]+/g)) {
+      let start = match.index,
+        chunk = match[0],
+        end = start + chunk.length,
+        parsed = Number.parseInt(chunk, 10),
+        face,
+        invalid = false;
+      if (Number.isFinite(parsed) && parsed >= 1 && parsed <= 9) face = String(parsed);
+      else if (Number.isFinite(parsed) && parsed >= 10 && parsed <= 15) face = "ABCDEF"[parsed - 10];
+      else if (Number.isFinite(parsed) && parsed === 16) face = "0";
+      else { face = chunk.toUpperCase(); invalid = true; }
+      entries.push({ face, start, end, invalid });
+    }
+    return entries;
+  }
   while (index < text.length) {
     let character = String.fromCodePoint(text.codePointAt(index));
     if (/[\s,;|]/.test(character)) {
@@ -1908,7 +1928,7 @@ function hodlDPlusRolls(value, targetWords = Pt, numberedD16 = hodlDPlusNumbered
     rejectedD8 = 0,
     rejectedD16 = 0,
     acceptedCharacters = [];
-  entries = hodlDPlusTokens(value);
+  entries = hodlDPlusTokens(value, numberedD16);
   let rolledEntries = entries.slice(0, rolledCharacterTarget),
     wordSlots = Array(rolledTarget).fill(""),
     groups = [],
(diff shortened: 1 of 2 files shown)


w-s-bitcoin:
Closing as superseded by #97, which has now merged with all checks passing.

This PR correctly identified that multi-character decimal D16 results could not be tokenized character-by-character. However, it fixes a decimal-input mode that no longer exists after #97. The merged implementation instead follows the published D++/Keysa workflow directly: D8 faces `1–8`, followed by two hexadecimal D16 faces `0–F`. It removes the decimal/hex selector and its saved state, so there is no decimal tokenizer or sanitizer path left to repair.

That also avoids codifying the separate `1 → 1, …, 16 → 0` convention, which does not express the published hexadecimal die faces directly. #97 locks the documented endpoints into CI: `100 → abandon` and `8FF → zoo`, and the complete GitHub test suite—including Firefox integration—passed.

Thank you for isolating the tokenizer failure and adding focused coverage. The underlying safety issue is resolved, but merging this branch now would reintroduce an unsupported notation and conflict with the canonical implementation.

PR #111, opened 2026-08-29
Offer explicit brain-wallet trimming conventions for recovery (#94)

### LINKED ISSUE #94: Brain-wallet recovery trims passphrases used by bitaddress.org v2.9.6+
## Summary

Brain-wallet recovery unconditionally trims leading and trailing whitespace before SHA-256. That is compatible with early bitaddress.org releases, but bitaddress.org deliberately removed this trimming in version 2.9.6 and hashes the entered JavaScript string as UTF-8 without boundary normalization.

EntropyLab advertises "bitaddress.org-style" recovery without selecting a source version. A passphrase created by bitaddress.org 2.9.6 or later with leading or trailing whitespace therefore derives a different private key and different addresses in EntropyLab.

Security review status: identified through static source review at commit `853d53db21cecf6abb935a7ae578ff813a67d794` (version 0.1.3). This has not been validated with an executable proof of concept. Any validation should use synthetic phrases and testnet or regtest only.

This is an EntropyLab compatibility issue, not a vulnerability in bitaddress.org.

## Affected Code

- `src/js/app.js:314-318` assigns `a = e.trim()` before the brain-wallet branch and hashes `a`.
- `src/js/app.js:705-710` advertises old passphrase-wallet recovery without identifying the trimming convention.
- `src/js/app.js:759` passes the textarea value into that path.

Upstream compatibility references:

- bitaddress.org v2.8.1 `src/ninja.brainwallet.js` trims with `.replace(/^\s+|\s+$/g, "")` before hashing.
- bitaddress.org commit `0456012aa0a58c96189e45e1adeb1a8df574082a` deliberately removes that trimming.
- bitaddress.org v2.9.6 and later hash the raw DOM string value.

## Recovery Impact

An honest user can enter the exact visible passphrase from a v2.9.6+ wallet and receive unrelated addresses. The original funds remain controlled by the historical key, but the advertised recovery workflow can make the backup appear wrong or empty.

The reverse compatibility problem also matters: always preserving whitespace would disagree with older bitaddress.org versions that trimmed it. A version-aware recovery choice is safer than silently applying either convention.

## Synthetic Reproduction

1. Use the synthetic phrase ` recovery phrase `, including both boundary spaces.
2. Compute SHA-256 over its exact UTF-8 bytes, matching bitaddress.org v2.9.6+.
3. Compute SHA-256 over `recovery phrase`, matching older trimmed behavior and current EntropyLab.
4. Confirm that EntropyLab selects the second key without disclosing the normalization.
5. Repeat with a leading/trailing tab, newline, and an all-whitespace synthetic phrase.

## Recommended Fix

Offer explicit recovery conventions, for example:

- `bitaddress.org 2.9.6+ (exact text)`
- `legacy bitaddress.org (trim boundary whitespace)`

Display a prominent warning when boundary whitespace is present and show which convention is being applied without altering the input silently. Do not expose or log the phrase.

## Acceptance Criteria

- Exact-text mode matches bitaddress.org v2.9.6+ vectors with leading/trailing spaces, tabs, and newlines.
- Legacy mode matches pre-2.9.6 trimmed vectors.
- The selected convention is visible before derivation.
- Boundary whitespace is never removed without disclosure.
- Existing WIF, hex, and Mini-key whitespace handling remains unchanged.

portlandhodl:
## Problem

Brain-wallet recovery unconditionally trimmed leading/trailing whitespace before SHA-256. That matches early bitaddress.org, but bitaddress.org deliberately removed trimming in **v2.9.6** and now hashes the entered string as UTF-8 without boundary normalization. A passphrase created by bitaddress.org 2.9.6+ with boundary whitespace therefore derived a different private key and different addresses in EntropyLab, with no disclosure of the normalization.

## Fix

The Brain wallet format now shows an explicit hashing-convention picker (visible only when Brain wallet is selected):

- **bitaddress.org 2.9.6+ (exact text)** — the default; SHA-256 of the passphrase exactly as entered, whitespace included.
- **Legacy bitaddress.org (before 2.9.6)** — trim boundary whitespace before hashing.

Disclosure, per the issue:

- The live status line names the active convention before derivation, and whenever boundary whitespace is present it says whether it is **kept** or **trimmed**.
- The derivation notes in the wallet result record which convention was applied.
- Blank passphrases are still rejected; WIF, hex, and Mini-key whitespace handling is unchanged (asserted by test).

## Tests (`test/brain-wallet.test.mjs`, added to `test:ci`)

- Exact mode returns the raw text (spaces, tabs, newlines) and its SHA-256 matches hashing the exact bytes.
- Legacy mode trims boundary whitespace and preserves interior whitespace; its SHA-256 matches hashing the trimmed text.
- Both conventions reject blank passphrases; an unknown convention value can never silently trim.
- Source assertions pin the default, the picker markup, the disclosure strings, and the brain-only visibility toggle.

## Verification

- `npm run build`
- `npm run test:ci` (306 pass)

Closes #94.

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
     "build:wasm": "node scripts/build-wasm.mjs",
     "clean": "node scripts/build.mjs --clean",
     "test": "node --test test/*.test.mjs",
-    "test:ci": "node --test test/network-check.test.mjs test/browser-check.test.mjs test/ui-defaults.test.mjs test/validate.test.mjs test/descriptor.test.mjs test/seedqr.test.mjs test/cards.test.mjs test/copy-seed.test.mjs test/number-bases.test.mjs test/dice-fairness.test.mjs test/seed-numbers.test.mjs test/dplus.test.mjs test/psbt-low-r.test.mjs test/psbt-metadata.test.mjs test/psbt-sighash.test.mjs test/secret-clear.test.mjs test/sqlite-writer.test.mjs test/wallet-export.test.mjs test/psbt-anti-exfil.test.mjs test/address-match.test.mjs test/psbt-nonce-reuse.test.mjs test/lifehash.test.mjs test/secp256k1-wasm.test.mjs",
+    "test:ci": "node --test test/network-check.test.mjs test/browser-check.test.mjs test/ui-defaults.test.mjs test/validate.test.mjs test/descriptor.test.mjs test/seedqr.test.mjs test/cards.test.mjs test/copy-seed.test.mjs test/number-bases.test.mjs test/dice-fairness.test.mjs test/seed-numbers.test.mjs test/dplus.test.mjs test/psbt-low-r.test.mjs test/psbt-metadata.test.mjs test/psbt-sighash.test.mjs test/secret-clear.test.mjs test/sqlite-writer.test.mjs test/wallet-export.test.mjs test/psbt-anti-exfil.test.mjs test/address-match.test.mjs test/psbt-nonce-reuse.test.mjs test/lifehash.test.mjs test/secp256k1-wasm.test.mjs test/brain-wallet.test.mjs",
     "test:dice-fairness": "node --test test/dice-fairness.test.mjs",
     "test:network": "node --test test/network-check.test.mjs",
     "test:browser-check": "node --test test/browser-check.test.mjs",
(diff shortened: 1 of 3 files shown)


w-s-bitcoin:
Thanks for the work here. We are taking a smaller UI direction for #94: an off-by-default trim checkbox beside the keyboard, generic recovery wording, and per-key state. That also preserves all-whitespace input in exact mode. Superseded by #136, so I am closing this version.

PR #118, opened 2026-08-29
Seed autocomplete defaults, D++ coin-flip keypad, and BIP87 multisig accounts

dplusplus1024:
OOGA BOOGA. Coin flip only show two rock. Passphrase box hide until you make word.

### DIFF
--- a/src/css/styles.css
+++ b/src/css/styles.css
@@ -435,6 +435,9 @@ label.field { display: block; margin-top: var(--space-component); font-size: 14p
 }
 .seed-word-meta:has(.muted:empty) { display: none; }
 .seed-word-meta > .muted { flex: 1 1 auto; min-width: 0; margin: 0; }
+/* Status lines quote what was typed, so a long unbroken token must break
+   rather than run off the side. */
+.seed-word-meta > .muted, #seed-meta, #dice-meta { overflow-wrap: anywhere; }
 .seed-word-copy-row {
   display: flex; align-items: center; justify-content: flex-end; gap: 8px;
   margin-top: var(--space-component);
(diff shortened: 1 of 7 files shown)


w-s-bitcoin:
I like seeing the passphrase xfp box before anything is entered there in the same way that we see the base xfp box before the seed is fully entered.

dplusplus1024:
> I like seeing the passphrase xfp box before anything is entered there in the same way that we see the base xfp box before the seed is fully entered.

I actually really like it being hidden, because passphrases are optional, while base seeds aren't. I think having the blank passphrase box is confusing to users, because it looks like something is broken or missing if they don't enter a passphrase.

dplusplus1024:
Took the hidden passphrase xfp box out — I still like it hidden, but it's your call and it's gone.

Rebased on the latest `rock` while I was in there. One addition: Taproot multisig now defaults to purpose 87 instead of 86. BIP86 is a single-sig standard, so it doesn't really say anything about a multisig co-signer key, and BIP87 is the account standard for exactly that. The Purpose field is still free, so anyone can type a different path.

Worth flagging that this needed a second change: purpose 87 used to be reachable only via the Legacy BIP87 checkbox, so any `87h` origin was treated as evidence of Legacy P2SH. With Taproot on the same purpose, selecting Taproot and pasting valid `87h/0h/0h` keys silently reselected Legacy and derived P2SH addresses with no warning. A BIP87 account key is script-agnostic, so it no longer picks a script type at all.

w-s-bitcoin:
Technically don’t they all actually have a passphrase? 😅😂

dplusplus1024:
> Technically don’t they all actually have a passphrase? 😅😂

I guess, if you consider an empty string to be a passphrase?

PR #120, opened 2026-08-29
Add scalable address range derivation UX

w-s-bitcoin:
## Summary

- replace fixed address-count presets with starting-index and range fields for singlesig and multisig
- add device-based time estimates and determinate progress with clean Stop cancellation
- render large address ranges with continuous virtual scrolling and contained table scroll behavior
- keep the selected starting address as the displayed receive QR

## Testing

- npm run ci (306 tests passed)


portlandhodl:
Rebasing

PR #127, opened 2026-08-29
Inspect raw signed transactions and label session-key output ownership.

MrHodlX:
PSBT / Nonce now accepts a fully signed raw Bitcoin transaction (hex or base64) as well as PSBT v0. With a session seed, root xprv, WIF, or hex key it labels each output as receive, change, or not in this wallet (accounts 0-2, 50 receive + 50 change, four script types). Two or more outputs with none ours is a destination-swap warning. Detector only: no chain, no node, no minting.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,13 @@ Official website: [entropylab.online](https://entropylab.online)
   Every input's declared sighash policy and each signature's appended sighash
   byte are decoded without a key; anything other than exact SIGHASH_ALL is a
   blocking warning.
+- Accepts a fully signed raw Bitcoin transaction (hex or base64) in the same
+  inspector: outputs, extracted ECDSA nonces, and inscription-envelope hints.
+  Fee and RFC 6979 cannot be checked without previous outputs.
+- With a session seed, root xprv, WIF, or hex key, labels each output as
+  change, receive, or not in this wallet (accounts 0–2, 50 receive + 50
+  change, all four script types). A two-or-more-output transaction with no
+  matching change is a blocking warning.
 - Runs a quick barrage of startup sanity checks on the host browser (secure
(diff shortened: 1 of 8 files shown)


J25dunn:
Rechecked this against current rock. The implementation has already landed indirectly through stacked PR #131: current rock contains commit 277df4b, which is the same raw-transaction/ownership change as this PR's 7c859c5. src/js/tx.js and src/js/ownership.js are byte-identical; test/tx-ownership.test.mjs differs only in the later Address import alias.\n\nVerified on current rock (60c2741):\n- node --test test/tx-ownership.test.mjs: 13/13 passed\n- npm run build: passed\n- npm run verify: passed\n\nNo duplicate merge is needed; this PR can be closed as superseded by #131 / commit 277df4b.

portlandhodl:
Closing as superseded: this change already landed on rock via stacked PR #131 (commit 277df4b), as verified in the comment above and confirmed in review.

Review of this branch found no money-losing issues — the parser and ownership logic check out against published BIP44/49/84/86 vectors, real signed transactions, and fuzzing. Two findings from the review remain live on rock and are worth a follow-up issue (not blocking here):

1. hodlDeclaredOutput still returns "lie" ("PSBT lies … Do not sign") for a legitimate multisig change output when the session key is a cosigner: the pubkey at the declared path matches, but the address check only tries the four singlesig script types, so a P2WSH/P2SH-multisig script can never verify.
2. hodlOwnershipWarning consults only the 50-address scan map, not the declared-output verification — so an honest change output at index >= 50 renders "change … (verified)" and the red "No output belongs to this session wallet" banner on the same screen. Any >= 2-output multisig spend with a session key loaded also always trips the destination-swap banner.

The ordering-dependence bug in this branch's hodlDeclaredOutput (first-entry short-circuit) is already fixed on rock (issue #194).

PR #132, opened 2026-08-30
Fail closed on below-threshold card transcripts (#85)

### LINKED ISSUE #85: Card entropy mode permits derivation from trivially enumerable transcripts
## Summary

The Cards input mode accepts any non-empty valid card transcript as wallet entropy. Falling below the selected BIP39 entropy target produces only a warning, so a single card can still produce a copyable 24-word phrase and a mainnet wallet.

This is not fixed by hashing the transcript: SHA-256 expands the representation, but a one-card input still has only 52 possible values. An observer who knows or suspects this workflow can enumerate every candidate wallet and match one against a public address, xpub, or transaction.

## Affected Code

- `src/js/app.js:1253-1262` calculates the required count and entropy estimate.
- `src/js/app.js:1305-1316` rejects empty/invalid/duplicate input but treats a below-threshold count as a warning and returns `ok: true`.
- `test/cards.test.mjs:120-126` explicitly asserts that one card returns a valid 32-byte entropy result with only a warning.
- `src/index.html:54,60-64` exposes Cards alongside normal key modes and defaults to a 24-word phrase.

## Security Impact

For a one-card transcript, an attacker has only 52 candidate inputs to test. For each card, the attacker can reproduce the canonical transcript hash, BIP39 mnemonic, and standard BIP32 addresses. Matching any public wallet data identifies the victim's mnemonic and private keys, allowing theft of every UTXO controlled by that wallet.

The same root cause affects every accepted below-threshold card count. Practical exploitability depends on the actual number of cards and whether an attacker can infer that this method was used.

## Preconditions

- A user enters fewer cards than the entropy target requires.
- The user proceeds despite the testing warning and funds the resulting wallet.
- An attacker can observe a public address, xpub, or transaction and can estimate a tractable input length.

## Synthetic Reproduction

1. Select Cards and a 24-word seed.
2. Enter one valid card, for example `AS`.
3. Observe that the application warns about `1 of 58` cards but still treats the entropy result as valid and permits wallet derivation.
4. Independently enumerate all 52 one-card transcripts and derive their synthetic testnet addresses.
5. Confirm that the address derived from `AS` uniquely identifies its generated mnemonic among those candidates.

## Recommended Fix

Fail closed until `hodlCardWithoutReplacementBits(cardCount)` reaches the selected BIP39 entropy target. While below the threshold, disable seed copying, wallet derivation, xprv/private-key output, and mainnet-compatible export.

If short deterministic vectors are needed for testing, isolate them behind an explicit test mode that forces testnet and cannot export production-compatible secrets.

## Acceptance Criteria

- One-card and all other below-threshold transcripts cannot produce or export a usable wallet.
- Threshold boundaries are tested for 12-, 18-, and 24-word phrases.
- The required counts are derived from the selected entropy target rather than duplicated as independent policy constants.
- Complete valid card transcripts continue to derive the expected deterministic test vectors.

FNBIP:
## Summary

Revives #109, which was closed without merging and without review comments. The vulnerability described in #85 is still live at current `rock` HEAD: `hodlCardsEntropy` pushes only a warning for a below-threshold transcript and still returns `ok: true` with derived key bytes, so a single card (52 enumerable candidates) produces a copyable 24-word phrase and a mainnet wallet.

Both original commits are cherry-picked from `fix/cards-entropy-threshold` with authorship preserved (`-x` annotated). They applied cleanly onto current `rock`, including the Ian Coleman suit-symbol mode and the direct rank-draw mode that landed after #109 was written.

## What it does

- **Fail closed at the single chokepoint**: below the required card count, `hodlCardsEntropy` returns `ok: false` with a blocking error instead of a warning. Every downstream output — derivation, seed copying, xprv/private-key output, wallet export, master-fingerprint preview — already gates on `entropy.ok`, so one gate blocks them all. (The direct rank-draw mode already fails closed on incomplete transcripts; only the hashed mode was affected.)
- **Required counts derived, not hardcoded**: `hodlCardNeeded` computes the smallest without-replacement deal whose entropy reaches the selected BIP39 target via `hodlCardWithoutReplacementBits`, per the issue's acceptance criteria. Values are unchanged: 25/31/39/50 cards, or 52+6 across two shuffles for 24 words, so complete-transcript test vectors are untouched.
- **Preview stays, copy locks**: the provisional word grid still renders during entry, but its copy control stays disabled below the threshold and the status line reads "N more required before deriving".
- **SECURITY.md** documents the new fail-closed posture for hashed card transcripts.

## Tests

- One-card transcript returns `ok: false` with no `bytes` (replaces the test that asserted the vulnerable behavior).
- Exact boundary tests for every phrase length: `required − 1` cards fail, `required` succeeds (12/15/18/21/24 words).
- A complete 58-card transcript still derives the expected deterministic SHA-256 seed and a checksum-valid 24-word mnemonic.

`npm test` passes 325/325; `npm run ci` (test + build + site verification) also passes.

Closes #85.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -38,14 +38,16 @@ material. Its security posture rests on the following model:
 - Wallet security depends on the quality and secrecy of the entropy, seed
   phrase, passphrase, or private key supplied by the user, and on the
   integrity of the machine it runs on.
-- Low-entropy dice and card transcripts are accepted intentionally so the
-  calculator can be used for deterministic tests, demonstrations, and
-  recovery experiments. EntropyLab does not claim that hashing a short input
-  makes it secure. When the entered transcript is below the recommended
-  entropy target, the result displays a prominent warning with the estimated
-  supplied entropy and says to use it only for testing. Users who intend to
-  secure funds must meet the displayed roll/card recommendation and verify
-  their procedure independently.
+- Low-entropy dice transcripts are accepted intentionally so the calculator
+  can be used for deterministic tests, demonstrations, and recovery
+  experiments; the result displays a prominent warning with the estimated
+  supplied entropy and says to use it only for testing. Hashed card
+  transcripts instead fail closed: below the card count required for the
+  selected phrase length, derivation, seed copying, and wallet export stay
+  disabled, because a short card transcript is trivially enumerable and
+  hashing a short input does not make it secure. Users who intend to secure
+  funds must meet the displayed roll/card requirement and verify their
+  procedure independently.
 - The single-file design inlines all scripts (`script-src 'unsafe-inline'`),
   and the secp256k1 WebAssembly module adds `wasm-unsafe-eval` to the
   content security policy: Chromium and WebKit engines refuse to compile a
--- a/src/js/app.js
+++ b/src/js/app.js
@@ -3026,12 +3026,14 @@ function hodlBinaryEntropy(value, targetWords = Pt) {
   return hodlNumberBaseEntropy(value, "bin", targetWords);
 }
 function hodlCardNeeded(targetWords = Pt) {
+  // Derived from the selected BIP39 entropy target, not policy constants:
+  // the smallest without-replacement deal whose entropy reaches the target.
+  // One deck tops out at ~225.6 bits, so a 256-bit seed finishes with extra
+  // cards from a second shuffled deck.
   let bits = hodlSeedConfig(targetWords).bits;
-  if (bits <= 128) return { first: 25, extra: 0 };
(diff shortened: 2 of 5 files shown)


w-s-bitcoin:
Thanks for the thorough work. EntropyLab intentionally allows low-entropy hashed inputs for testing, recovery, and deterministic experiments, with a clear warning rather than a hard block. Since this PR changes that product policy, we won’t merge it. The derived recommendation counts and complete-transcript vector test could still be useful in a separate non-blocking PR.

PR #138, opened 2026-08-30
Trim the beta banner copy and make it dismissible

timechainb:
The banner led with "Beta software:" and carried a second sentence about securing real bitcoin. Drop that sentence, set the label on its own line in the header's alert red, and reword the remainder to "EntropyLab is experimental and currently should only be used for testing."

Add a dismiss control in a narrow column at the banner's top right. The choice is remembered in localStorage, keyed to the build version so every release warns again, and wrapped so a storage-less origin (file://, private modes) simply shows the banner on every load. The modal disclaimer still gates the page and still carries the funds warning.

The uppercase label is scoped to .beta-warning-text rather than .beta-warning, which the noscript notice also uses and which would otherwise have uppercased "JavaScript is required:" too. #beta-warning restates [hidden] because its own display: flex outranks the user agent rule.

Claude-Session: https://claude.ai/code/session_01Le3Hsku2vLe7iqM8RepxJq


w-s-bitcoin:
I did not merge [PR 138](https://github.com/w-s-bitcoin/entropylab/pull/138).
The functionality is otherwise sound and all 323 tests passed against current rock, but the dismiss icon is forced white in both themes. In light mode it has only about 1.33:1 contrast against the pale banner, making it difficult to see.
Recommended fix:
- Use color: var(--fg) for the dismiss icon.
- Update the browser test to verify theme-aware visibility rather than requiring white.
PR 138 remains open. Local rock is clean and unchanged.

PR #144, opened 2026-08-30
Show Bitcoin Core exports for SLIP-132 imports

### LINKED ISSUE #129: Show Core xpub and descriptor next to an imported ypub/zpub
Overheard conversation from D++ and Wicked about various pubs..
Tried to understand the conversation with my clanker Grok.. came up with an idea. 

## Problem

Ongoing compatibility issue between SLIP-132, used by Electrum 
and some hardware wallets, which stamps script type into the 
extended-key prefix (`ypub`, `zpub`, `Ypub`, `Zpub`), and 
Bitcoin Core's Output Descriptors, which represent script type 
explicitly in the descriptor and use plain `xpub`/`tpub` keys.

EntropyLab already accepts those prefixes and already builds Core
descriptors. The easy recovery mistake is still: paste a `zpub`,
then import the same key as the wrong script (`pkh()` instead of
`wpkh()`, or singlesig instead of multisig). Watch-only is empty;
coins are on the other script.

## Suggestion

On a successful SLIP-132 import, print these three lines in one
place on the result / recovery pane:

1. What was pasted — `zpub…` (or `ypub` / `Ypub` / `Zpub` / testnet twin)
2. Same key as Core expects — `xpub…` or `tpub…` (payload unchanged,
   version bytes rewritten)
3. Descriptor to give Core — e.g.
   `wpkh([fpr/84h/0h/0h]xpub…/<0;1>/*)#checksum`

If the stamped prefix and the selected script type disagree, show a
high-visibility warning before derive.

Do not add a tab, a prefix-flip tool, or extra default export
formats. Keep wallet.dat / descriptor output as it is.

### Example

For example, after importing a `zpub`, the recovery pane could show:

```text
Imported key:
zpub6...

Core xpub:
xpub6...

Descriptor:
wpkh([fpr/84h/0h/0h]xpub6.../<0;1>/*)#checksum
```

If the user instead selects `pkh()`:

```text
⚠️ Script type mismatch

Imported key: zpub6...
Indicates:    wpkh(...)
Selected:     pkh(...)

This may derive different addresses from the imported key.
```

The same could apply to other SLIP-132 prefixes, e.g. `ypub → sh(wpkh(...))`.

This makes it immediately clear what the imported key represents, how it is represented in Bitcoin Core, and which descriptor should be used to verify the expected addresses.

## Checks

Published Electrum / SLIP-132 vectors work for this
(abandon…about → known account `zpub` and first `bc1q` addresses).

If the recovery sheet already shows (2) and (3) on import, this can
be closed.

(just a noob trying to help - please feel free to ignore if me/my clanker do not make sense!)

Jdelg718:
## Summary

- group an imported single-signature SLIP-132 key, its equivalent Bitcoin Core `xpub`/`tpub`, and a checksummed Core multipath descriptor in one recovery block
- warn prominently before derivation when a `ypub`/`zpub` (or testnet twin) conflicts with the selected script type, while continuing to derive the script encoded by the prefix
- keep generic `xpub`/`tpub` imports script-agnostic and avoid fabricating master fingerprints or origin paths
- integrate with configurable derivation hardening and branch selection from current `rock`; impossible hardened public-key paths still fail before any script warning, while the recovery descriptor remains the conventional `<0;1>/*` receive/change descriptor

## Safety

- conversion changes only the four version bytes; the serialized key payload remains identical
- the grouped export contains public/watch-only data only and uses the existing escaped field renderer
- no network egress, entropy generation, dependencies, or generated artifacts added

## Tests

- `npm run ci` — 472/472 passed; build and artifact verification passed
- `node --test test/slip132-import.test.mjs` — 6/6 passed
- official BIP84 zpub/xpub/address vector and independently checked Core descriptors
- mainnet/testnet ypub/zpub-family payload identity, descriptor wrappers, and checksums
- configurable branch regression confirms Core recovery stays on `<0;1>/*`
- generic xpub exclusion, hardening precedence, and script-mismatch behavior
- Firefox UI flow verifies the warning and grouped recovery export
- independent fail-closed review of the rebased conflict resolution

`npm test` additionally passed 501/502 locally; the sole failure was the Firefox integration launcher because Firefox is not installed on this machine. GitHub CI is the Firefox/browser gate.

Closes #129.


J25dunn (commented):
Reviewed the SLIP-132 conversion, descriptor construction/checksum, mismatch warning, and output rendering. I found no defect in the submitted commit: conversion changes only the four version bytes, all four single-signature y/z mainnet/testnet families preserve the 74-byte payload, the published BIP84 zpub maps to the expected xpub/address, and the Core multipath descriptor checksum is correct. Unknown origins remain intentionally omitted and rendered values use the existing escaped field helper.\n\nIndependent verification on a0874ea:\n- node --test test/slip132-import.test.mjs: 5/5 passed\n- npm run ci: 329/329 passed; build and artifact verification passed\n- git diff --check: passed\n\nThe branch is now behind current rock (60c2741) and conflicts in package.json and src/js/app.js after the configurable-derivation merge. The app.js conflict overlaps imported-key hardening checks/status and descriptor branch rendering, so this should be rebased and the integrated result rerun before merge. Happy to re-check the resolved revision.

Jdelg718:
@J25dunn Thanks for the careful review. I rebased this onto current `rock` (`02a7a2c`) and resolved the configurable-derivation conflicts in `package.json` and `src/js/app.js`.

The resolved revision preserves the new hardening/branch validation, keeps the SLIP-132 mismatch warning for valid imports, and constructs the recovery descriptor independently as the conventional `<0;1>/*` Core receive/change descriptor even when the UI derives a custom branch.

Verification on `d9bd487`:
- `node --test test/slip132-import.test.mjs`: 6/6 passed
- `npm run ci`: 472/472 passed; build and artifact verification passed
- GitHub Firefox integration: passed
- independent fail-closed rebase review: passed
- `git diff --check`: passed

The PR is now clean and mergeable. Happy to have you re-check the resolved revision.

portlandhodl:
Superseded by #211, which merges this PR with the overlapping #208 into one branch. Your commit is preserved verbatim in that branch's history (and credited via Co-authored-by on the integration commit): the script-mismatch warning and the Bitcoin Core recovery export are both in. One integration note: the recovery block now renders only the conventional Core descriptor, because #208's prefix-swap fields directly above it already show the pasted key and its Core equivalent — showing all three twice read as a duplication bug in the combined branch. Thanks @Jdelg718!

PR #145, opened 2026-08-30
Add offline iPhone home-screen app

### LINKED ISSUE #137: FIX APPLE PHONE NO NET
iOS no open local html file. Must use PWA magic. Put code inside web box so heavy fruit phone save page forever. Even when internet die, app open quick from PWA cache.

Jdelg718:
## Summary

- add a manifest and Apple home-screen metadata so iPhone Safari can install EntropyLab as a standalone web app
- cache only the self-contained hosted entry points for offline Home Screen launches, with production-only registration and no network fallback
- derive the service-worker/cache identity from the built HTML, manifest, and worker template so updates cannot drift while the package version stays unchanged
- document the iPhone workflow and preserve the downloaded, verified HTML as the recommended path for sensitive use

## Security model

- registration is restricted to HTTPS on `entropylab.online` or `www.entropylab.online`; downloaded `file://` HTML never registers a worker
- the worker handles only same-origin GET navigations from its current named cache and has no `fetch()`, push, sync, or notification behavior
- precaching bypasses stale HTTP-cache bodies; a failed precache never activates or deletes the prior usable cache
- generated HTML and worker IDs are verified equal before publication
- `connect-src 'none'` remains unchanged; no entropy generation or dependency was added

## Tests

- `node --test test/pwa.test.mjs` — 8/8 passed
- `npm run ci` — 332/332 passed; build and artifact verification passed
- `npm test` — 338/339 passed locally; sole failure is the Firefox launcher because Firefox is not installed on this machine. GitHub CI is the Firefox/browser gate.
- independent fail-closed security/spec review

Closes #137.


portlandhodl:
Superseded by #204 (rebased onto current rock, conflicts resolved, CI green) and merged there. Thanks @Jdelg718 — authorship was preserved on the rebased commit.

PR #149, opened 2026-08-30
Add a stacked-deck backup of a derived seed

ottoz0r:
## Summary

- encode a derived BIP39 seed as a unique without-replacement card order (mixed-radix, least-significant digit first; same deal counts as hashed cards: 25 / 31 / 39 / 50 / 52+6)
- restore it from Cards only after an explicit action: pick the derived key and click Show stacked deck (or Restore a stacked-deck backup on this tab). Switching tabs never loads a backup. A nonempty transcript requires Replace the cards currently on this tab
- hashed and direct dealing stay the main Cards flow. The backup row sits above the transcript, compact, with the note "Turn an existing seed into a card-order backup"
- Show stacked deck also turns on Show cards so the faces appear with the transcript
- this is not a SHA-256 hashed card transcript and will not match Ian Coleman method. 

## Open question

Should **Card deck backup** stay on the Wallet data private recovery section?

It is gated on Show private recovery material (same as SeedQR), then collapsed behind a details summary. The feature is niche: seed to stacked-deck backup. It may be better to drop it from the result page and keep only Cards → Show stacked deck.

## Safety

- deterministic encoding of user-supplied BIP39 entropy; no generated randomness
- no network egress
- result-page backup is omitted from the DOM until private recovery is shown; the recovery sheet only writes CARD DECK BACKUP when that box is on
- the derived-key dropdown lists name, method, and word count, not the seed

## Tests

- `node --test test/cards.test.mjs test/ui-defaults.test.mjs` after rebase onto current `rock`
- round-trip encode/decode for every BIP39 length; hashed transcripts are not valid backups; incomplete and out-of-range deals are rejected

Generated `entropylab.html` is not in this PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,10 @@ Official website: [entropylab.online](https://entropylab.online)
   separate as well; conventional script choices restore their standard purpose,
   while pasted co-signer origins auto-detect and must agree with the selected
   purpose.
+- Encodes a derived BIP39 seed as a unique without-replacement card order
+  (Card deck backup). On Cards, that order is shown only after you select the
+  derived key and click Show stacked deck. It is not a hashed card transcript
+  and will not match Ian Coleman.
 - Inspects PSBT v0 transactions, reports PSBT-provided amounts and fees, checks
   for repeated ECDSA nonces from the same public key, verifies optional Jade
   anti-exfil (sign-to-contract) transcripts without a key, and can compare supported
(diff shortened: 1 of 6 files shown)


ottoz0r:
Closing it for now. Might be useful later as a physical backup of a seed you already derived (not a hashed shuffle), but today it would be useful only for EntropyLab and might be a footgun

PR #152, opened 2026-08-30
Parse PSBT v2 (BIP370) into the existing inspector.

MrHodlX:
PSBT inspector currently dies on v2 files. This adds helpers that turn a BIP370 v2 PSBT into the same unsigned-tx shape the v0 inspector already walks.

- src/js/psbt-v2.js
- test/psbt-v2.test.mjs
- Inspect only. No signing.

Run: node --test test/psbt-v2.test.mjs


MrHodlX:
Opened on the wrong branch. Replacing with a clean PR that only has the two v2 files.

PR #163, opened 2026-08-30
Treat browser online status as unverified

### LINKED ISSUE #158: Offline mode not detected in Tails
You probably already have this in your backlog but just in case, opening v0.1.3 on Tails (offline) still shows the online badge and note.

J25dunn:
Closes #158.

Tails uses Tor Browser, and current Firefox anti-fingerprinting can deliberately force `navigator.onLine` to `true`. That makes passive detection unable to distinguish an offline Tails session from a connected browser. EntropyLab also cannot safely resolve the ambiguity with a connectivity probe because its runtime must remain network-silent.

This changes the header status semantics:

- `navigator.onLine === true` is shown as **Unverified**, not **Online**
- `false` remains **Offline**, with accessible text explaining that it is only the browser report
- the initial markup ships in the cautionary **Unverified** state
- the stale planned "Network detected" modal copy is removed
- README and SECURITY document the limitation
- tests cover initial, online-event, offline-event, and Network Information API transitions

The red caution state and `connect-src 'none'` policy remain intact. No network probe, dependency, application cryptography, or wallet derivation behavior is added or changed.

References:

- Mozilla documents `navigator.onLine` as inherently unreliable: https://developer.mozilla.org/en-US/docs/Web/API/Navigator/onLine
- Firefox anti-fingerprinting forces it true: https://bugzilla.mozilla.org/show_bug.cgi?id=1975851

Validated with:

- `npm run ci`
- `npm run test:network`
- `npm run test:ui`
- `npm run test:validate`
- `git diff --check`

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,6 +86,13 @@ modern browser. For sensitive wallet material, use a dedicated air-gapped
 machine and verify important addresses and descriptors with an independent
 wallet or signing device before receiving funds.
 
+The header status is only a browser hint. Some privacy-focused browsers,
+including Tor Browser configurations, deliberately report `navigator.onLine`
+as true even when networking is disabled. EntropyLab therefore labels that
+state **Unverified** and never probes the network. An **Offline** label means
+only that the browser reported offline; verify the air gap yourself before
+entering wallet secrets.
+
 To build the HTML file yourself, see [Building from source](#building-from-source).
 
 ### Verifying the download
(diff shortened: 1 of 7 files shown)


portlandhodl:
Approach nACK, Should always show online if there is any doubt.  Would suggest that we completely overhaul this section to get a true or false. 

Option:

user can select from a dropdown of servers to go a GET command and check? Must be user initiated.

Issue #169, opened 2026-08-30
add shell: bash explicitly to CI - ensure windows availability


RandyMcMillan:
reference: https://github.com/RandyMcMillan/entropylab/actions/runs/33329486663/job/99305452178

RandyMcMillan:
testing now: https://github.com/RandyMcMillan/entropylab/actions/runs/33335657319/job/99322004608

portlandhodl:
Reviewed against current `rock` (`1bd67d0`): **still applicable**. `.github/workflows/ci-cd.yml` contains no `shell:` keys at all (15 `runs-on` jobs rely on the default shell per platform). Open PR #177 (".github/workflows/ci-cd.yml: add shell: bash") is the pending fix.

PR #177, opened 2026-08-30
.github/workflows/ci-cd.yml:add shell: bash

RandyMcMillan:
00-00000261


RandyMcMillan:
fixes: #169

https://github.com/RandyMcMillan/entropylab/actions/runs/33335657319/job/99322004608

SHA256SUMS.txt
RandyMcMillan, line 1: check sums need to be verified before passing CI

PR #178, opened 2026-08-30
Add an i18n foundation with English keys and initial Spanish, Portuguese, French, and German translations

ottoz0r:
UI copy is extracted into English translation keys instead of hardcoded strings. English is the source of truth catalog. New languages are extra JSON files with the same keys, not app rewrites.

This also matters when the hosted page is shared. The document language follows the active locale, so a link opened in Spanish, Portuguese, French, or German is marked and shown in that language instead of always presenting as English. That helps browsers, assistive tech, and search treat the page as the language the reader is actually using.

How the foundation works:
- Catalogs live in `src/locales`. `en.json` defines every key. Other locales are complete only when they have the same keys and no empty strings.
- Runtime lookup is `hodlT` / `t`, with `{placeholder}` interpolation. Static markup uses `data-i18n`, `data-i18n-html`, `data-i18n-aria`, and `data-i18n-placeholder`.
- User-facing errors and notes store `{ key, vars }` via `hodlError` / `hodlNote`, so a locale change re-translates messages already on screen.
- `hodlApplyLocale` re-renders chrome, forms, results, PSBT copy, and derive/import/multisig errors from the active catalog.
- The header language selector only lists complete catalogs. The choice is remembered.

This PR also ships a first translation of the English keys into Spanish, Portuguese, French, and German. Those catalogs should get a further native-speaker pass; that review is follow-up, not a merge gate.

To keep this scalable:
- Add or change copy in `en.json` first, then the same key in every complete catalog.
- Do not concatenate sentences in JS. Put the full string in the catalog and interpolate with `{name}` placeholders.
- New user-facing errors and notes must be specs (`hodlError` / `hodlNote`), not English `throw new Error("...")`.
- Do not offer a language in the selector until its catalog matches English keys with no empty strings.


dani69654:
Lots of code for languages is not worth imo. English is enough.
The repo's goal is clear and we should not add code complexity for functionalities that are out of scope

portlandhodl:
Rebased onto the tip of `rock` (`3a68c97`) and completed the conversion gaps that opened up since this branched. Head is now `6a68285`.

**Rebase notes.** Rock meanwhile landed the identifier rename, the header network picker, the Key Station method/seed-length dropdowns, the multisig descriptor import, and the trailing-path parser. Conflicts were resolved by keeping rock's current code and re-applying the i18n conversion on top of it — a few strings the branch carried had regressed to older copy and were restored to rock's text.

**New conversions for post-base features**, all keyed and translated (es/pt/fr/de):
- Network picker: option names/descriptions, menu title, button aria-label, and the never-connects note. The picker re-renders on locale change.
- Multisig "Paste descriptor" panel (summary, label, help, button) and the co-signer session-key/derivation-path aria labels.
- Trailing-path parser errors (wildcard/multipath rules), hardened account-key import errors, seed word-number errors, brain-lab notes and acknowledgement errors, derive stop/stopping arias, "Show calculations" toggles, PSBT session fallbacks, and the import script-mismatch/detection notes.

**Catalog fixes:**
- Synced stale entries to current copy (`action.derive` is "Derive Key", the longer `msig.intro`, `msig.legacyBip87Help`, `error.import.rootXpubPaths`, `error.msig.cosignerDuplicate`, …).
- Fixed a factual error present in all five catalogs: the D++ coin read is 1–4 Heads / 5–8 Tails (the catalogs had it inverted).

**Latent issues fixed while merging:**
- `formatLabel`/`formatShort`/`formatUnit` were referenced but never defined — the Number bases mode would have thrown.
- Note/error specs would have rendered as `[object Object]` in the wallet safety notes, the brain-lab preview, and several entropy/validation failure paths; `hodlWalletMessages` is spec-aware now and a `hodlThrowIfFailed` helper rethrows specs as translated `hodlError`s.
- Key and multisig tab state persists the error spec, so errors restore translated when switching tabs.
- `hodlApplyLocale` handles rock's dropdowns (select option refresh) and the network picker.

**Housekeeping:** `entropylab.html` is deliberately not committed — rock's convention is that feature commits don't touch the artifact and CI rebuilds it.

**Verified:** `npm run build`, `npm test` (872 pass / 0 fail / 1 pre-existing skip), `npm run test:browser` green in headless Firefox and Chrome, and a headless smoke test of the built file booting in Spanish (`<html lang="es">`, translated header, filled five-option selector).

PR #198, opened 2026-08-31
Verify every PSBT output derivation claim, not just the first

### LINKED ISSUE #194: Security: PSBT output derivation ordering can suppress the “PSBT lies” warning
## Summary

`hodlDeclaredOutput()` returns after examining the first `PSBT_OUT_BIP32_DERIVATION` entry in an output map.

Because an output map may contain multiple derivation entries, and their serialized order is controlled by the PSBT creator, reordering otherwise identical entries changes EntropyLab’s security verdict.

A foreign-fingerprint entry placed first prevents EntropyLab from examining a later entry that uses the loaded wallet’s fingerprint but does not actually produce the output. The output is classified as belonging to another wallet instead of displaying the intended **“PSBT lies … Do not sign”** warning.

Affected code: `src/js/app.js`, function `hodlDeclaredOutput()`.

## Relevant code

```js
for (let entry of hodlFind(entries, 2)) {
  // ...
  if (fp !== fingerprint)
    return { state: "other-wallet", path: label, fingerprint: fp };

  // Matching-fingerprint verification...
}
```

The first foreign fingerprint causes an immediate return, so later derivation records are never checked.

## Reproduction

1. Load a root key as the PSBT session key.
2. Construct a PSBT with at least one genuine wallet output, such as legitimate change. This prevents the global “No output belongs to this session wallet” warning.
3. Give another output two `PSBT_OUT_BIP32_DERIVATION` records:

   - First: a foreign master fingerprint.
   - Second: the loaded wallet’s fingerprint, but with a key/path that does not produce the output script.

4. Inspect the PSBT.

With the foreign record first, `hodlDeclaredOutput()` returns:

```json
{
  "state": "other-wallet"
}
```

The later false claim involving the loaded wallet is not examined, so the per-output **“PSBT lies”** warning is absent.

Reverse the same two records and the function returns:

```json
{
  "state": "lie"
}
```

The security verdict therefore depends only on map-entry order.

## Impact

An attacker-controlled PSBT can suppress a warning intended to detect false change or ownership metadata.

The global ownership warning does not compensate when another output genuinely belongs to the session wallet: `hodlOwnershipWarning()` stops after finding any wallet-owned output. The affected output remains visible, but the explicit metadata-integrity warning is missing.

Multiple derivation records are also normal for multisignature outputs, so returning after the first record is not a safe assumption.

## Expected behavior

The result must be invariant under permutation of the output’s derivation records.

EntropyLab should:

1. Examine every derivation record.
2. Ignore foreign fingerprints while looking for records claiming the loaded wallet.
3. Validate every applicable matching-fingerprint claim.
4. Return `lie` if a record claims the loaded wallet but its pubkey, path, or output script does not verify.
5. Return `other-wallet` only when no record claims the loaded wallet.
6. Handle legitimate multisignature derivation records without making the result depend on cosigner order.

## Suggested regression tests

Add tests covering:

- A foreign derivation followed by an invalid matching-fingerprint derivation.
- The same records in reverse order.
- A foreign derivation followed by a valid matching-fingerprint derivation.
- Multiple legitimate multisig cosigner derivations in every permutation.
- A transaction containing genuine change plus another output with a false matching-fingerprint claim.

Every permutation of the same records should produce the same verdict.

portlandhodl:
Closes #194.

## Problem

`hodlDeclaredOutput()` returned `other-wallet` upon encountering the first foreign-fingerprint `PSBT_OUT_BIP32_DERIVATION` record in an output map. Since an output map may carry several derivation records (normal for multisig cosigners) and the PSBT creator controls their serialized order, a foreign record placed first hid a later false claim naming the session wallet: the per-output **"PSBT lies ... Do not sign"** warning never appeared. Reversing the same two records produced `lie` — the security verdict depended on attacker-controlled ordering.

The global ownership warning does not compensate: `hodlOwnershipWarning()` goes quiet as soon as any output (e.g. genuine change) belongs to the session wallet.

## Fix

`hodlDeclaredOutput()` now scans **every** derivation record:

- Foreign fingerprints are skipped, remembering the first as the `other-wallet` fallback.
- Every record claiming the session wallet's fingerprint is verified (pubkey vs. derived path, and output script vs. derived addresses).
- Any failing claim returns `lie` regardless of position; a verifying claim yields `ours`; `other-wallet` is returned only when no record claims the wallet.

The verdict now depends on the set of records, not their order.

## Verification

New `test/psbt-output-derivation.test.mjs` (wired into `test:ci`) covers every case from the issue:

- false matching-fingerprint claim (wrong keydata, or right keydata but wrong output script) behind a foreign record, in **both orders** → `lie`
- valid claim behind a foreign record, both orders → `ours`
- every matching claim validated even after one verifies
- multisig cosigner records in all permutations → order-invariant verdict
- genuine change present + false claim on another output → still `lie`
- malformed records skipped; no session key → `null`

Confirmed the new suite fails 6/9 tests against unpatched `origin/rock` code and passes 9/9 with the fix. Full `npm run build && npm test` passes (733 tests, 0 failures).


J25dunn (changes requested):
The ordering attack is fixed and the focused suite passes, but there is one blocking false-positive in the new multisig case.\n\nThe test named 'multisig cosigner records give the same verdict in every permutation' constructs a legitimate 2-of-2 P2WSH output and then adds a valid derivation record for the loaded session root. It explicitly expects every permutation to return . That verdict reaches hodlRenderOutputHtml as **'PSBT lies ... Do not sign'**, even though a PSBT_OUT_BIP32_DERIVATION record in a multisig output claims that this derived pubkey participates in the output script; it does not claim that the pubkey alone encodes the P2WSH output. Comparing that pubkey against only p2pkh/p2sh-p2wpkh/p2wpkh/p2tr single-key templates therefore cannot prove a lie.\n\nPlease either verify the matching pubkey against the committed redeem/witness script (and verify that script against the output scriptPubKey), or return an explicit neutral/unsupported result for matching-fingerprint multisig claims rather than . The regression should assert that legitimate multisig cosigner records remain order-invariant without producing the 'PSBT lies' warning.\n\nIndependent checks on 5c877fc:\n- focused suite: 9/9 passed\n- build and artifact verification: passed\n- order-suppression cases and all-matching-record validation otherwise look correct.

J25dunn:
Correction to my review: the two omitted code-formatted words are the state name 'lie'. The legitimate multisig test currently expects state 'lie', and the requested neutral/unsupported alternative should replace 'lie' when the implementation cannot verify multisig script participation.

portlandhodl:
Thanks for the review.

PR #201, opened 2026-08-31
fix: imported non-root extended keys are reported as the root xpub

dani69654:
## Summary

Fixes a js comma operator typo in `Po()` (`src/js/app.js`) that caused an account-level (non root) extended key to be labeled as the BIP32 root xpub.

### The bug

```js
// before (comma operator — depth check ignored)
rootXpub: (s.depth === 0, s.publicExtendedKey),

// after (ternary — matches rootXprv on the line above)
rootXpub: s.depth === 0 ? s.publicExtendedKey : null,
```

Importing a depth-3 account xprv/xpub produced a non-null wallet.rootXpub, so the recovery sheet and UI labeled it "BIP32 ROOT XPUB" even though the app itself flags it as non-root.

### Fix: 
One-line ternary, matches rootXprv and the existing depth-0 pattern at app.js:429.

### Test:
 npm run build && npm test; manual import of account-level keys now yields rootXpub: null.


portlandhodl:
Thanks for the careful read of the extended-key import path — the comma operator on that line is a real bug *in the source text*, and the ternary matches the `rootXprv` guard beside it.

However, while preparing a regression test I found that this line is **unreachable at runtime**. `src/js/app.js` carries two generations of the wallet pipeline:

- old declarations: `uf` (line 305), `mf` (423), `Hs` (427), `Po` (455 — the function this PR patches), `Oo` (503)
- the new pipeline, assigned at module load and replacing those bindings: `uf = function` (1211), `mf = function` (1392), `Hs = function` (1419), **`Po = function` (1428)**, `Oo = function` (2074), `tc = function` (2056)

The only caller of `Po` (line 1142) resolves the binding at call time, so it always invokes the new `Po`, which already returns `rootXpub: null` for account-level imports (lines 1448–1449), with script-type detection and a single-account result. Nothing captures the old function object before the reassignment, so the old `Po` — comma bug included — never executes.

Verified empirically (full untouched `app.js` loaded as a module under Node with DOM stubs, real `HDKey`, importing a depth-3 `m/84'/0'/0'` account xpub):

- base (`7862bc3`, without this fix): `rootXpub: null`, one detected account
- this PR: identical result

The same dead-code status applies to the old recovery-sheet builder (`Oo`), so the "BIP32 ROOT XPUB" mislabeling described above can't be produced from the current base. A telling confirmation: the deployed bundle still contains the comma expression — esbuild minifies the discarded `s.depth === 0` down to `s.depth` — alongside the new `Po` that actually runs. Both the buggy line and its replacement were introduced in the same commit (`9353390`), so the bug has never been live.

Where this leaves the PR: the change is a correct repair of a latent bug in shipped-but-unreachable source — zero risk, but also zero runtime effect — and the premise (user-visible mislabeling) doesn't reproduce on the current base. Per CONTRIBUTING there's no changed behavior to pin with a test (any behavioral test passes on the base too). Possible directions:

1. Reframe this PR as dead-code correctness (update the description) and merge as a latent-bug fix, or
2. Close it and instead remove the dead old pipeline (`uf`/`mf`/`Hs`/`Po`/`Oo`/`tc` declarations plus the `To`/`cr` trim at lines 1182–1194), which eliminates this whole class of confusion, or
3. Land a behavioral test for the *live* contract either way: importing an account-level key yields `rootXpub: null` and no root-xpub label anywhere in the output.

Happy to help with 2 or 3 if there's interest.

---

*Postscript: this PR has been merged as a zero-risk latent-bug fix; the analysis above is for the record and to inform a possible follow-up that removes the dead pipeline outright.*

dani69654:
Thanks for reviewing @portlandhodl , didn't notice it was not used tbh :)

PR #203, opened 2026-08-31
Opinionated changes to design system: restyle, workspace rail, and project banners

zmeyer44:
Some opinionated styling changes to get feedback on. No worries if this isn't the style you guys are looking for, but wanted to share it anyway
_Users_claudius_entropylab_entropylab html

## 1. `769355d` — Restyle onto Geist

The stylesheet gains a token layer: Geist's published colour scales, declared step for step per theme under their own names, with the app's existing semantic names (`--bg`, `--surface`, `--fg`, `--border`, `--accent`, …) resolved onto them. Nothing downstream was renamed, so the change is a value change an auditor can follow one hop at a time.

- Primary action is Geist's inverted button rather than orange. The global momentary press feedback stays but moves against each control's own ground via `--press-bg/-fg/-border`. Orange survives as the brand — the logo, and the die-distribution bars.
- One focus ring everywhere, replacing twenty-two ad-hoc outlines in five colours.
- Radii collapse to Geist's two materials, 6px and 12px, from the ten values in use.
- Geist type roles: published sizes, leading and negative tracking; body at copy-14; nothing below the 12px step. Headings move off the serif — Geist sets everything in one sans.
- Fifty `color-mix()` computations become the designed tint steps; ad-hoc shadows become Geist's tooltip/menu/modal shadows.
- Anti-patterns Geist names by name: the all-caps letter-spaced eyebrow, the bevelled slider knob, the three-stop progress gradient, the glow under the threshold fill.

**No webfont.** Geist Sans/Mono would have to be fetched or embedded and the tool must not touch the network, so the platform faces stand in at Geist's metrics. Two further deliberate departures, both stated in the stylesheet: controls take step 5 for their border rather than step 4 (a 1.2:1 edge on white), and a tinted button's edge is its own text colour for the same reason.

### Pre-existing bugs fixed on the way

- Printing from the dark theme put the dark theme's chroma on paper — the print block redefined nine greys but not `--accent`, `--danger`, `--ok` or `--warn`. It now restates the Geist light scale.
- Dropdown options cancelled their focus outline, leaving a 1.2:1 wash as the only cue. Every `<select>` in the app is one of these, including the 2048-entry BIP39 word list.
- The multisig tab icon was three fixed greys, near-invisible on the light tab strip.

### The rail

Above 1280px the workspace switcher becomes a rail pinned to the viewport's left edge under the header. Below it, the hamburger and its panel are unchanged.

**A sidebar was built here once (166dc46) and reverted (d7d5bec)** for widening the page to 1260px and giving up its single measure; the test at the end of the workspace suite forbade its return by name. That decision has been reversed deliberately, and the rail answers the original objection rather than repeating it: it is fixed against the viewport instead of wrapped around the content, and the content column keeps the page's own 1000px measure.

The breakpoint is arithmetic, not taste — the rail may only appear once the page can pay for it without leaving the column narrower than the 968px it has with no rail:

| viewport | mode | content column |
|---|---|---|
| 1024 | dropdown | 968 |
| 1279 | dropdown | 968 |
| **1280** | **rail** | **993** |
| 1288+ | rail | **1000** |

Monotonic: widening the window never narrows the tool. A test does that subtraction, so moving the breakpoint or the rail's width fails with the numbers in the message.

One breakpoint owns the layout and the open state — the media query and `hodlWorkspaceRailWidth` hold the same number, held together by a test — and a `matchMedia` listener closes the dropdown on the way up so `aria-expanded` never describes a hidden control. The query is scoped to `screen`, or a wide print box would keep the hidden rail's gutter as a blank margin.

### The banners

The beta and hosted-site warnings become Project Banners: icon, one sentence in sentence case, and the control that resolves the state. Severity moves out of the wording into the variant — online is `error`, a beta build is `warning` — so the two stop competing in the same red.

They are **not dismissible**, which is the component's defining rule: a state that can be cleared without being resolved belongs in a Note. That removes both close controls, the two `localStorage` keys, the pre-paint `data-beta-banner-dismissed` attribute and its head-script check, and `hodlInitBetaWarningDismiss`.

The blocking "Beta software / I understand" modal is **unchanged** — Geist puts confirmations in a Modal, and it is a deliberate click-through gate for a tool that handles keys.

## Tests

`npm test` 726 · `npm run test:browser` 50 Chrome checks · `npm run verify` reproducible.

Rebased onto current `rock` (`23a3133`). Six files conflicted; upstream had independently made the online banner non-dismissible, which this branch converges with, and had replaced the derivation-path preview that the restyle had touched — the deleted component was taken, not restyled. Two rules that landed on `rock` after the sweep (`.derivation-advanced summary`) were brought into the system to match their siblings.

Nine assertions were retired and replaced, each because the behaviour it described is gone, and each replacement stated in place with its reason:

- the orange press-feedback contract, the progress bar and slider handle that no longer carry gradients, and the kicker that gave up its capitals;
- the two-theme token check, which now follows the token indirection one hop instead of matching a hex;
- the dismiss controls and stored dismissals — replaced by their inverse, so re-adding a close control to a safety warning has to be deliberate;
- the guard forbidding the sidebar, replaced by the rail's own contract.

Nine new assertions cover the print block's parity with the light theme, the multisig icon's tokens, the tab strips' hover opt-out, the rail's geometry and breakpoint arithmetic, and — in the browser suite, because no source-text assertion can see it — that an address row really renders at the height the virtual scroller assumes.

**Every assertion added here was checked by mutation**, not by reading. Two rounds of adversarial review ran against the diff; between them they confirmed 24 findings, including four regressions I had introduced and three of my own tests that passed vacuously. All are fixed.


portlandhodl:
I actually dig this quite a bit, one larger issue is the removal of orange elements from the dark mode theme replacing them with white elements.  This breaks Entropy Lab branding.

portlandhodl:
@timechainb

PR #208, opened 2026-08-31
Show SLIP-132 extended keys as a prefix swap only.

MrHodlX:
Import/derive lists the key as pasted, the Bitcoin Core xprv/xpub (or tprv/tpub), then a matching y/z form when path and script agree.

- Same payload, new version bytes and checksum.
- Script lives in the descriptor, not the prefix.
- x = legacy, y = nested BIP49, z = native BIP84. Testnet: t / u / v. No Taproot SLIP prefix.
- A generic xprv is not re-prefixed unless the pasted prefix already matches.

Independent of #144 (Core export grouping).

Run: `node --test test/slip132-display.test.mjs`


portlandhodl:
Superseded by #211, which merges this PR with the overlapping #144 into one branch. Your commit is preserved verbatim in that branch's history (and credited via Co-authored-by on the integration commit): the prefix-swap display, purpose-gated SLIP-132 family, and docs are all in. One integration note: the non-progress import path now also passes importedFamily/importedValue — without it, the two import paths resolved different primary families, which disabled the Core recovery export on one of them. Thanks @MrHodlX!

PR #209, opened 2026-08-31
Add Brain wallet — lab: SHA-256 text as 256-bit BIP39 entropy.

MrHodlX:
Hashes exact UTF-8 text with SHA-256 and uses the 32-byte digest as 256-bit BIP39 entropy (24 words).

- Private-key brain wallets remain a separate scalar path.
- Not a BIP39 passphrase, not a Bitcoin Core hdseed or address-key backup.
- Derive Wallet is required. The mnemonic is not previewed while typing.
- Strength is the entropy of the text, not the word count. Guessable text is stolen coins.
- Session ack stays in page RAM.

Run: `node --test test/brain-wallet.test.mjs`

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,13 @@ Official website: [entropylab.online](https://entropylab.online)
 - Accepts dice rolls, coin flips, hexadecimal entropy, BIP39 seed phrases,
   extended keys, WIF keys, raw private keys, and Casascius mini private keys.
   All five BIP39 phrase lengths (12, 15, 18, 21, and 24 words) are supported
-  for every entropy entry method.
+  for every entropy entry method. A separate **Brain wallet — lab** mode hashes
+  exact UTF-8 text with SHA-256 and uses the 32-byte digest as 256-bit BIP39
+  entropy (24 words). That is not a BIP39 passphrase, not a Bitcoin Core hdseed
+  or address-key backup, and not the private-key brain-wallet path (which treats
+  the same hash as a secp256k1 scalar). Strength is the entropy of the text, not
+  the word count. Derive Wallet is required; the lab does not preview the
+  mnemonic while typing.
 - Derives BIP39 seeds, BIP32 extended keys, wallet fingerprints, addresses,
   and Bitcoin Core-compatible descriptors. Each master fingerprint is shown
   next to its deterministic [LifeHash](https://lifehash.info) icon so two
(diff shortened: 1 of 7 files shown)


portlandhodl:
ACK 68824d3f4c1bb9b440ab6194a3d8f15cbb78718d

Reviewed and tested locally (security / mergability / correctness scan):

**Correctness**
- Verified the mapping end-to-end against an independent from-spec BIP39 implementation: SHA-256("correct horse battery staple") = c4bbcb1f…e39a8a → same 24 words from the app's `entropyToMnemonic` and my independent encoder. Trailing whitespace changes the digest, matching the "Whitespace is significant" copy.
- Integration is complete: tab array/labels and the index.html placeholder stay in sync (live DOM shows exactly 6 tabs), derive gating requires the session ack plus non-empty text, the derive path re-validates and throws otherwise, seed-length control and master-fingerprint preview are hidden, `hodlFingerprintMnemonic` returns null (no silent mnemonic preview), and state capture/restore/clear all handle the new `brainLab` field.

**Security**
- In-policy per CONTRIBUTING §1: a deterministic transformation of user-supplied text, no generated entropy. No network egress added; the same-origin-connections browser check stays green. Single self-contained entropylab.html still builds reproducibly.
- No injection surface: user text only reaches `.value`/`textContent`, never `innerHTML`.
- Disclosure is thorough and honest — pre-use ack warnings, per-result warnings, and README/SECURITY.md/llms.txt all state this is not a BIP39 passphrase, not a Core hdseed/address-key backup, and not the same wallet as the private-key scalar brain path. Derive path wipes `entropy.bytes` in a finally.

**Mergability**
- `git merge-tree --write-tree origin/rock HEAD` merges cleanly, no conflicts. All CI checks pass.

**Commands run**
- `npm run build` (v0.1.3, reproducible)
- `npm test`: 744 pass / 0 fail / 1 skip (Edge not installed locally); Firefox and Chrome each pass all 49 browser integration checks
- `node --test test/brain-wallet.test.mjs`: 6/6 pass
- Live-DOM check of the built artifact (headless Chrome): 6 mode tabs, brain-lab flow renders as designed

Non-blocking observations for the record: (1) the live hex preview is information-theoretically the wallet — the words stay hidden until Derive, but the displayed SHA-256 hex converts trivially; the UI copy discloses what appears, and verifiability is arguably the point of a lab mode, but worth being deliberate about. (2) Per-keystroke digests in the preview path aren't zeroed (the derive path is); consistent with how other modes treat in-flight input. (3) The ack is session-global rather than per-key, matching the documented intent. (4) A browser-suite case clicking the new tab would close the E2E gap; the node test plus existing BIP39 vectors cover the behaviour today.

PR #212, opened 2026-08-31
Add global entropy synchronization

w-s-bitcoin:
## Summary
- replace the post-derivation workspace copier with opt-in live entropy synchronization
- keep one consistently labeled, always-enabled sync checkbox directly below the key-method selector
- synchronize direct dice, card, number-base, BIP39, WIF, and private-key hex representations live, truncating each destination to complete symbols
- let hashed inputs publish their derived entropy one way to non-hashed methods while never allowing non-hashed edits to overwrite hashed transcripts
- remove the number-base-only sync control and isolate BitBox input from hashed dice input
- preserve invalid source text and exact source WIF encodings
- rerun synchronization for on-screen pad clicks and card undo actions that do not emit ordinary input events

## Verification
- `npm run build`
- `npm run verify`
- `npm run test:browser` (52 Chrome integration checks passed; Firefox and Edge unavailable locally)
- `npm test` (763 passed, 2 unavailable-browser skips)


portlandhodl:
Rebasing + Reviewing

portlandhodl:
Rebased onto current `rock` (`618d5ae`), resolving all merge conflicts with the merged manual-calculation-education and psbted-actions work: kept rock's `showNumberBaseCalculations` toggle/restore/capture wiring and calculation styles, kept this PR's removal of workspace sync + number-base sync and the new global sync control. The rebased diff vs `rock` is exactly this PR's change set.

Also completed a full security audit of the sync implementation: bits↔transcript mappings verified exact in both directions against the reference parsers (D++ `(face−1)·256 + d16·16 + d16`, BitBox 5×2-bit+coin, direct-card mixed radix, number-base mixed-radix remainders); BIP39 checksum-candidate ordering confirmed so final picks equal entropy suffixes; hashed methods can never publish or be reverse-filled; no RNG/network/storage; wipe and pagehide paths reset all new state.

One functional gap found and fixed in `b42938f`: the dice pad (`[data-d]`), card undo (`#card-undo`), and last-word picks mutate their textarea without a bubbling `input` event, so the sync did not re-run on pad-only entry until the next keystroke. The delegated click handler now covers them (seed-length buttons deliberately excluded — re-syncing there could clear destinations when the active method has nothing to publish). Covered by a new node source-guard test and a new browser-suite test.

Verification: `npm run build` ✅, `npm run verify` ✅ (byte-reproducible), `npm test` ✅ 764/764 tracked tests, `npm run test:browser` ✅ Firefox + Chrome (Edge absent, skipped by design).

PR #221, opened 2026-09-01
Gate the brain wallet, warn on short synced entropy, show paths with apostrophes, and tidy the key and PSBT action rows

dplusplus1024:
Follow-up to #215, plus two fixes found while looking at it.

### The brain wallet stays behind its acknowledgement

Selecting Brain wallet now shows only the output choice, the warning and the checkbox. The input, its label, help, trim toggle, keyboard and status appear once the warning is acknowledged, and hide again if it is cleared. Other private-key formats are untouched.

The acknowledgement is **per output**. The two outputs carry different warnings — the seed-phrase one adds two bullets about mnemonics — so acknowledging the single key pair no longer unlocks the seed phrase, or the other way round. Each remembers its own answer for the session, so switching between them does not ask twice.

The warning heading reads "read before use" rather than "read before first use", since it is meant to be re-read. The derive error no longer refers to "the lab warning", which named a mode that no longer exists.

### Short synced entropy is a caution, not a tick

The badge reported every amount as a green tick, including keys built from far
too little entropy. Syncing itself is unchanged — everything still propagates
exactly as before — but the badge now says whether what went in was enough.

The first attempt measured the published bit string, which for any hashed
method is the 256-bit SHA-256 output however few rolls produced it, so a single
dice roll still read "256 bits synced" in green. It now reads the source
estimate the app already shows beside the input, the same figure as
"2.6 bits estimated", and takes the smaller of that and the published count, so
a digest can never look stronger than the rolls behind it. Brain wallets report
no estimate, since their strength is the text rather than a countable input, so
they do not get a false green either.

The healthy case reads "Key synced" rather than a bit count, because the count
was overstating things by construction. Below 128 bits the tick stays, since it
did sync, and the shortfall is named beside it:

```
  1 roll     Key synced   2 bits of entropy under 128
 20 rolls    Key synced  51 bits of entropy under 128
102 rolls    Key synced
```

The threshold is `hodlGlobalSyncMinimumBits()` rather than a literal.

On a narrow screen the badge drops below the toggle and the shortfall takes its
own line, while each phrase stays intact.

The first attempt at that widened the badge at every width, which squashed the
sync toggle on desktop. The desktop rules are now exactly what they were, and
the wrapping lives in the existing 720px breakpoint. Measured at both widths,
with no horizontal overflow at either:

| | toggle | badge |
| --- | --- | --- |
| 1280px | 598px | 286px, same line, text on one line |
| 375px | full width | below the toggle, shortfall on its own line |

### The PSBT editor's action row

The network select is a plain `Bitcoin mainnet` again, with Testnet chosen from the list.

The row now lines up. Three things stacked: the select was 44px beside 36px buttons, its label kept the 16px field margin because `label.field` outranks the row's own rule, and the whitespace either side of the control added another 8px of line boxes. The label lays out as flex so whitespace-only text is ignored, drops the margin, and the select and buttons share one height.

```
            height   top   bottom
Load PSBT       36   884      920
Clear editor    36   884      920
Network         36   884      920
```

That also closed the dead space under the PSBT input: the row went from 68px to 44px, leaving a normal 16px gap.

### Layout, everywhere these pieces appear

- The derivation progress pill matches the 44px control height of the buttons beside it, and takes a 1px edge from `currentColor`, so it is accent while deriving and the same sage green as its label once complete.
- The on-screen keyboard sits with the field it types into, above the master fingerprint cards, rather than stranded between them and the derivation settings.
- The keyboard toggle keeps its square size when the passphrase options grow to two rows, instead of stretching to match them.
- The BIP87 multisig option spells its path `m/87'/coin'/account'` with apostrophes rather than the `h` suffix, in that label only.

Both `src/index.html` and `src/js/app.js` carry these sections and the runtime renders the app.js copy, so each markup change is applied to both.

### A brain wallet reports unknown strength, not a green tick

Review caught that a brain wallet published the 256-bit length of its SHA-256
digest as though that were counted entropy, so any text at all — `password`
included — earned a clean green tick with no caution. Its strength is however
guessable the text is, which this app cannot measure.

This needs a third state rather than reusing `null`. Number bases and typed
seeds return `null` too, and there the published count genuinely *is* the
entropy — folding the two together would have warned on every healthy direct
method. So brain wallets get their own answer:

```
dice / cards      counted source estimate
private key       256, except a brain wallet
brain wallet      unknown
number bases      null, the published count is the entropy
seed phrase       null, the published count is the entropy
```

```
password          Key synced   entropy unknown, only as strong as the text
```

### A derived brain wallet does not outlive its acknowledgement

Also from review: unticking "I understand" hid the input but left the derived
wallet on screen and still revealable, which is the opposite of what the
checkbox promises. It now retracts the result through the app's own
`hodlInvalidateLiveKeyResult` path, clearing the render, the stored result and
the reveal flag.

Switching output does the same. The two outputs build different wallets from
the same text, so a result left over from the other one is a wrong address
waiting to be copied — it was staying on screen labelled "single-key wallet
data" while the HD radio was selected.

Verified end to end: derive from `correct horse battery staple`, reveal,
untick — the WIF and the digest are gone from the page, not merely hidden. The
scalar address matches an independent computation
(`bc1q08alc0e5ua69scxhvyma568nvguqccrv4cc9n4`).

One thing worth flagging that is **not** from this branch: deriving a brain
wallet leaves the key-kind radio on WIF, so a second derivation needs the kind
re-selected. It reproduces with these commits reverted, and this branch only
reads that radio, never renders it.

### The browser suite follows the new wording

`test/browser.test.mjs` is not part of `npm run test:ci`, so the wording change
looked green locally and failed the Firefox job. Four assertions pinned
`N bits synced`. Two of them aborted their test before its cleanup ran, which
left the sync toggle checked and the dice mode on D++ for the two tests that
followed — which is why a dice fairness test that has nothing to do with
syncing was among the failures.

The healthy cases now assert the tick with no shortfall element. Nothing is
dropped: the exact count for 255 bits is still pinned by the hexadecimal and
Base32 round-trip immediately below it, and the badge's own count-printing by
the 3-bit and 11-bit shortfall cases.

### Paths read with apostrophes, descriptors keep h

Paths shown to the reader now use the apostrophe form BIP32 itself writes, so
the address under the QR code, the receive and change tables, the Purpose
labels and the text recovery sheet all read `m/84'/0'/0'` instead of
`m/84h/0h/0h`.

Descriptors are untouched. Their key origins still emit `h`, which is what
descriptors and the wallets reading them expect, so the two conversions now run
in opposite directions on purpose. Typed input is unaffected either way, since
a derivation path already accepts both markers. Confirmed on one derivation:

```
display     m/84'/0'/0'/0/0
descriptor  wpkh([cd34af7b/84h/0h/0h]xpub.../0/*)#t40lyg7c
```

Several suite assertions that pinned the displayed form were awaited rather
than immediate, so each stale one cost the run fifteen seconds and pushed the
whole suite past its sixty-second budget — which surfaced as a timeout with no
test reports at all rather than as a failure. The descriptor-origin assertions
sitting beside them are deliberately left alone, so those tests now check both
conventions at once.

### The BIP-85 shortcut goes

The button beside Derive Wallet dated from when BIP-85 had no home of its own.
It has a tab now, so the shortcut was a second door to the same room, and the
key action row reads better as Derive, progress, Clear.

Only the shortcut and its click handler go. The BIP-85 tab keeps its own
"Use active key" button, which calls the same `hodlUseActiveKeyForBip85`
helper, so adopting the active key as a BIP-85 parent is unchanged. The test
that pinned the shortcut now pins its absence and additionally asserts the
shared helper and the tab's own button survived, so the removal cannot quietly
take the underlying feature with it.

### Checks

`npm test` 822 passing, 0 failing, 7 skipped, which includes the browser suite in Firefox, Chrome and Edge. `npm run build` and `npm run verify` clean. Rebased on rock.

Every change was confirmed in the browser by measuring what actually rendered rather than by eye: control heights, edges and border radius for the action rows, document order for the keyboard, the badge's text and warning element at one roll, twenty rolls and a full transcript, and the brain wallet's unknown state from a live derivation. The retraction was checked by deriving, revealing, then unticking, and confirming the WIF and the digest are gone from the page rather than merely hidden.

New behaviour is covered rather than assumed: the badge's three states by
source assertions, and the retraction by a browser test that derives, reveals,
revokes and checks the recovery material is actually gone from the page. That
test looks the acknowledgement checkbox up again before each click — it is
re-rendered on every refresh, and holding a reference clicks a detached node
and silently proves nothing.

Tests that pinned the old behaviour were updated rather than dropped — the keyboard's position assertion, the tools row's `align-items`, the PSBT row's shared height, and the brain-wallet invariants — so each still locks something down.

No change to the built artifact in this branch.


dani69654:
Tested carefully, lgtm

portlandhodl:
Review pass during merge prep (maintainer). The branch is now rebased onto current `rock` and three commits were pushed on top; the PR's own design and commits are untouched.

**Merge resolution (`41e4689`)** — rock had moved and the PR was conflicting:
- Rock removed the PSBT editor's network select outright (the editor follows the header network picker now), so the action-row restyle of that control had nothing left to restyle — rock's markup/CSS taken, and rock's tests already pin the select's absence.
- Rock independently removed the Derive BIP-85 child shortcut; the tests were folded so both sides' assertions survive (row order incl. `derive-progress`, no `bip85-open` listener, `hodlPickBip85SessionKey`/`hodlRefreshStationKeyPickers` pinned).
- Semantic conflict git can't see: rock's new Vanity-tab browser test pinned `m/84h/0h/3h` on a *display* surface while this branch converts display paths to apostrophes. Red in both engines; fixed to `m/84'/0'/3'/0/0`. Descriptor origins keep `h` everywhere, as this branch intends.

**Review fixes (`400df9e`)** — three gaps found in the new features, each with regression coverage:
- `hodlCloneDerivedKey` didn't carry the new `globalSyncSourceBits`, so committing a key (every derive) or editing inputs silently dropped the shortfall/unknown cautions and the badge fell back to digest length. The verdict now travels with the sync.
- Revoking the brain-wallet acknowledgement only cleared the *active* key. A brain wallet committed to another tab re-rendered in full — reveal included — with no acknowledgement in effect. Brain-derived results now carry a marker and revoking sweeps every slot (`hodlRetractBrainWalletResults`); the browser test now parks one brain wallet in its tab, revokes from a second, and checks the parked tab's material is gone.
- A minikey reported 256 bits of source entropy; the accepted 22-char form (S + 21 base58) tops out near 123 bits — under the feature's own 128-bit floor. It now reports the payload keyspace (58^n), matching how dice and cards are judged.
- Also dropped `hodlKeyModeLabels`, which was added here but never referenced.

Scope note on the sweep: it covers Key Station slots; a retracted key also drops out of every picker since they read `state.result`. BIP-85 children derived earlier are separate artifacts with their own wipe flow — flagging as a conscious boundary, not an oversight.

**Firefox CI (`1d43283`)** — the red job was the 60s report watchdog expiring, not a failing check: rock's Firefox already needs ~55s of that budget on the shared software-rendered runner, and the new revocation test tips it over. Watchdog raised to 150s with a comment that it detects hangs, not slowness.

Verified end to end on the pushed tree: `npm run build`, `npm run verify`, and `npm test` — 974 passing, 0 failing, including the browser suite (71 checks) in both Firefox and Chrome.

PR #225, opened 2026-09-01
Air-Gapped Raspberry Pi Build Guide - Alpine Linux

pdubl22:
Aloha!

I've put together a guide for creating a hardened, air-gapped, diskless Alpine Linux image that loads entropylab.html.

To keep the attack surface as small as possible, I've implemented a few layers of lockdown: disable network hardware at the firmware/kernel level, using a non-privileged user (entropylab) to ensure the Chromium Sandbox is fully active, and using a diskless (RAM-only) boot to ensure no persistence. I've added this as a separate guide file so it doesn't interfere with the main README. I hope this is a useful addition for the community.

I also tried making a change to the main entropylab.html to replace some lorem ipsum text, but it put it into this same pull request.

That doesn't seem right, but I don't know how to fix it!


portlandhodl:
Create a /docs folder and put this in there.

PR #240, opened 2026-09-01
Restore folder tabs for session wallets and keep LifeHash on rename

ottoz0r:
Reverts the chip tray from #230. Session wallets are folder tabs on the card again.

Double-click rename keeps the LifeHash. The old key silhouette is gone.


ottoz0r:
Folded into https://github.com/w-s-bitcoin/entropylab/pull/237 so we only rebase one branch.

PR #242, opened 2026-09-01
Update Single Signature wallet description from Lorem text

pdubl22:
New text:

Generate single-signature Bitcoin keys from your own custom entropy. Derive extended public keys, descriptors, and addresses across all standard script types, including native SegWit and Taproot. Verify your master fingerprints and export watch-only configurations to receive bitcoin without ever exposing your private key material to the internet.


dani69654:
ACK

PR #247, opened 2026-09-01
Fix duplicated card pad clicks

### LINKED ISSUE #243: Fix: card shuffle suit sometimes stays selected after a tap
## Summary

In the **Cards** input mode, tapping a rank (e.g. `A`) then a suit (e.g. `♥`) usually commits the card and resets the pad for the next one. Intermittently the suit button keeps its `.active` (orange) state and **no card is committed** — you have to tap the suit again. Reported by a tester on the live site.

## Root cause

A single physical tap dispatches **two** events to the pad handler: the touch/pointer event (`touchend` / `pointerup`) **and** the synthesized `click` the browser fires afterward. The commit handler ran twice on one tap:

1. First run → commits the card, clears both `.active` states, advances.
2. Second run → re-toggles the suit's `.active`, then no-ops the commit because no rank is selected anymore → **pad left half-selected, suit bold.**

## Proposed fix

- Route the pad through `click` only; `preventDefault()` on `touchend` / `pointerup` so the browser does **not** synthesize a second `click`.
- Clear `.active` and advance **synchronously** on commit, so the pad never rests in a half-selected state.
- Add a per-pad commit guard so a double-dispatched `click` is ignored.

## Repro

Cards mode → tap `A` → tap `♥`. Repeat ~20×.

- **Before:** roughly 1 in N taps leaves the suit bold and drops the card.
- **After:** every tap commits exactly one card and the pad resets.

## Notes

- No change to the entropy / hashing path — only the pad's event wiring.
- The built `entropylab.html` is generated; this edits the source module that wires `.card-suit-pad` / `.card-rank-pad`. Rebuild to regenerate the standalone file.

## Reference implementation

The names below are illustrative — match them to the real module's identifiers. The important parts are the three changes marked (1)(2)(3).

```js
// Reference fix for the EntropyLab Cards input pad.
//
// Drop this logic into the source module that wires `.card-suit-pad` and
// `.card-rank-pad` (the built entropylab.html is generated from that source).
// The names below are illustrative — match them to the real module's
// identifiers. The important parts are the three changes marked (1)(2)(3).

export function wireCardPad({ suitPad, rankPad, onCommit }) {
  // (3) Per-pad commit guard: a double-dispatched click is ignored so a single
  // physical tap never commits twice.
  let committing = false;

  let selectedSuit = null; // button element currently .active in suitPad
  let selectedRank = null; // button element currently .active in rankPad

  const setActive = (pad, button) => {
    pad.querySelectorAll("button").forEach((b) => b.classList.remove("active"));
    if (button) button.classList.add("active");
  };

  const clearSelection = () => {
    setActive(suitPad, null);
    setActive(rankPad, null);
    selectedSuit = null;
    selectedRank = null;
  };

  const tryCommit = () => {
    if (committing) return;          // (3) ignore a double-dispatched event
    if (!selectedSuit || !selectedRank) return; // need both halves of the card

    committing = true;
    try {
      // (2) Clear + advance synchronously so the pad never rests half-selected,
      // even if a stray second event lands before the (async) hash resolves.
      const suit = selectedSuit.dataset.suit;
      const rank = selectedRank.dataset.rank;
      clearSelection();
      onCommit({ suit, rank });       // onCommit may await hashing; that's fine
    } finally {
      committing = false;
    }
  };

  const handleSuit = (button) => {
    selectedSuit = button;
    setActive(suitPad, button);
    // Selecting a suit does not commit; wait for the rank. If a rank was
    // already chosen (user tapped rank first), commit now.
    if (selectedRank) tryCommit();
  };

  const handleRank = (button) => {
    selectedRank = button;
    setActive(rankPad, button);
    if (selectedSuit) tryCommit();
  };

  // (1) Route through click only and suppress the synthesized second event.
  // Without this, touchend/pointerup + click fire the handler twice per tap.
  const killSynthetic = (event) => event.preventDefault();

  suitPad.addEventListener("touchend", killSynthetic, { passive: false });
  rankPad.addEventListener("touchend", killSynthetic, { passive: false });
  suitPad.addEventListener("pointerup", killSynthetic);
  rankPad.addEventListener("pointerup", killSynthetic);

  suitPad.addEventListener("click", (event) => {
    const button = event.target.closest("button");
    if (button) handleSuit(button);
  });
  rankPad.addEventListener("click", (event) => {
    const button = event.target.closest("button");
    if (button) handleRank(button);
  });

  return { clearSelection };
}
```

Opening this as an issue rather than a PR since the reference fix uses illustrative names — happy to adapt it to the real source module and open a PR if that's preferred.

J25dunn:
Closes #243.

A card commit now keeps a per-render guard active through the current event turn, preventing a duplicated compatibility click from re-selecting the suit after the rank and suit have been cleared. The pads stay wired through `click` only, so touch, mouse, keyboard, and screen-reader activation are unchanged.

The browser regression dispatches the duplicated suit click and verifies that exactly one card is committed and neither pad retains an active selection.

Tests:
- `npm run build`
- `npm test` (829 passing)
- `npm run verify`
- Firefox and Chrome/Chromium browser suites (including the new regression test in both engines)

Edge was not installed.

### DIFF
--- a/src/js/app.js
+++ b/src/js/app.js
@@ -3583,7 +3583,7 @@ function hodlUpdateCards() {
   hodlCardRank = selection.rank;
   if (selection.card && !parsed.invalidRanges.length) {
     hodlCommitCardSelection(input, selection.card);
-    return;
+    return true;
   }
   let dealt = document.getElementById("dealt-cards");
   if (dealt) {
@@ -3632,6 +3632,7 @@ function hodlUpdateCards() {
   let undo = document.getElementById("card-undo");
   if (undo) undo.disabled = !parsed.cards.length && !String(input.value || "").trim();
   hodlQueueMasterFingerprintPreview();
+  return false;
 }
 function hodlSetInputValueAtEnd(input, value) {
   input.value = value;
@@ -5368,16 +5369,25 @@ function hodlRenderKeyForm() {
         hodlQueueMasterFingerprintPreview(0);
       };
     });
+    let cardPadCommitting = false;
     hodlFormEl.querySelectorAll("[data-card-suit]").forEach((button) => {
       button.onclick = () => {
+        if (cardPadCommitting) return;
         hodlCardSuit = hodlToggleCardChoice(hodlCardSuit, button.getAttribute("data-card-suit"));
-        hodlUpdateCards();
+        if (hodlUpdateCards()) {
+          cardPadCommitting = true;
+          queueMicrotask(() => cardPadCommitting = false);
+        }
       };
     });
     hodlFormEl.querySelectorAll("[data-card-rank]").forEach((button) => {
       button.onclick = () => {
+        if (cardPadCommitting) return;
         hodlCardRank = hodlToggleCardChoice(hodlCardRank, button.getAttribute("data-card-rank"));
-        hodlUpdateCards();
+        if (hodlUpdateCards()) {
+          cardPadCommitting = true;
+          queueMicrotask(() => cardPadCommitting = false);
+        }
       };
     });
     hodlBindKeypadPointer(hodlFormEl.querySelectorAll("[data-direct-card-rank], #card-undo"), () => input);
--- a/test/browser-suite.html
+++ b/test/browser-suite.html
@@ -432,6 +432,20 @@
     input(cards, "");
   });
 
+  await test("a duplicated card-pad click commits once and clears both selections", async () => {
+    await chooseMode("Cards");
+    const cards = await until(() => document.getElementById("cards"), "card transcript input");
+    const ace = document.querySelector('[data-card-rank="A"]');
+    const hearts = document.querySelector('[data-card-suit="H"]');
+    ace.click();
+    hearts.click();
+    hearts.click();
+    equal(cards.value, "Ah ");
+    assert(!document.querySelector(".card-suit-pad .active"), "Suit remained selected after the duplicated click");
+    assert(!document.querySelector(".card-rank-pad .active"), "Rank remained selected after the duplicated click");
+    input(cards, "");
+  });
+
   await test("partial entropy word grids remain copyable", async () => {
     await chooseMode("Number bases");
     document.querySelector('input[name="entropy-format"][value="bin"]').click();


portlandhodl:
Thanks for picking up #243 — the commit guard and the regression test are a good direction, and the change is nicely small. Unfortunately the `touchend`/`pointerup` `preventDefault()` half of the fix breaks the pads completely on touch devices, so this can't merge as-is.

**The problem:** the suit/rank pads' only activation handler is `onclick`. Per the Touch Events spec, canceling `touchend` suppresses the compatibility `click` for that tap — so on a touch device the tap produces *zero* clicks and the handler never runs. (Canceling `pointerup` is a no-op for click dispatch; it's `touchend` that kills it.)

**Verified empirically** (headless Chrome 152, trusted touch input via CDP `Input.dispatchTouchEvent`, mobile viewport emulation), against the real built `entropylab.html` from this branch vs. its base:

| Real tap | Base | This PR |
|---|---|---|
| rank `A` | `A` activates | nothing happens |
| suit `♥` | commits `Ah ` | nothing happens |

Isolated in a minimal repro with the exact wiring:

- `onclick` only → tap fires exactly one `click`, handler runs
- PR wiring (`touchend`+`pointerup` preventDefault on the pad) → `pointerdown/touchstart/pointerup/touchend` fire, **no `click`, handler never runs**
- `pointerup` preventDefault only → `click` still fires (no-op)
- `touchend` preventDefault only → **no `click`** ← the killer
- the codebase's existing `hodlBindKeypadPointer` pattern (`pointerdown` preventDefault) → `click` still fires

Two further checks on the built artifact:

- **Direct word selection is broken too** — the direct rank pad carries the `card-rank-pad` class, so the same listeners attach; a touch tap on a direct rank enters nothing.
- **Desktop mouse is unaffected** — real `Input.dispatchMouseEvent` clicks still commit `Ah `.

The new browser test passes either way because it dispatches `.click()` programmatically, which bypasses the touch pipeline entirely — the in-page suite can't observe this regression.

**Suggested fix:** drop the `touchend`/`pointerup` preventDefault block and keep the rest — the `hodlUpdateCards()` return value, the `cardPadCommitting` guard, and the regression test are a clean, low-risk hardening against a same-task double dispatch, and mouse/keyboard/screen-reader UX is unchanged. If you want extra consistency with the rest of the app, binding these pads through the existing `hodlBindKeypadPointer` (as the direct pad, undo button, and dice keypads already do) is safe — verified above that it still delivers exactly one click per tap — but it's optional and doesn't dedupe clicks by itself.

Worth re-testing on a real touch device (or CDP touch emulation) before re-requesting review; happy to re-run my harness on an update.

PR #251, opened 2026-09-01
Add offline Entropy Journal section

MrHodlX:
Encrypted notebook of user-supplied dice, coins, hex, brain-wallet text, or seeds. AES-256-GCM; the key is PBKDF2-SHA-256 (600,000 rounds) of a typed password, with the salt derived from the password itself and the IV derived as HMAC-SHA-256 of the plaintext — the file is a pure function of the password and the entries, so no CSPRNG is called anywhere. One downloaded JSON file. No localStorage, no relays, no invented entropy.

Rebased onto current rock: the Journal tab rides the locale system (workspace.journal / workspace.journalShort in all five catalogs).

Ran: npm run build, npm run verify, npm run test:ci (879 pass), npm run test:browser (64/64 Firefox, 64/64 Chrome).


portlandhodl:
Superseded by #282, which folds the encrypted notebook into the combined Journal tab (merged as 5e3f74f).

Issue #191, opened 2026-08-31
Add independent PSBT UTXO verification


Hero-Gamer:
## Motivation

The PSBT Editor https://github.com/w-s-bitcoin/entropylab/pull/167 now displays input amounts, scripts, and `WITNESS_UTXO` / `NON_WITNESS_UTXO` data, and can calculate fees. However, these remain **claims contained in the PSBT** rather than independently verified UTXO data.

For a verification-focused tool like EntropyLab, it would be useful to distinguish between:

> **“The PSBT says this UTXO is worth X”**

and:

> **“This UTXO actually matches the referenced transaction.”**

## Suggestion

Where the referenced previous transaction is available, verify each PSBT input's:

* transaction ID / outpoint
* output index
* amount
* scriptPubKey

For example:

```text
PSBT INPUT VERIFICATION

Input #0
  Outpoint:
    abc123...:1

  PSBT claims:
    Amount: 1.50000000 BTC
    Script: <scriptPubKey>

  Referenced transaction:
    Amount: 1.50000000 BTC
    Script: <scriptPubKey>

  ✓ UTXO verified
```

If the claim does not match:

```text
⚠️ UTXO MISMATCH

Input #1

PSBT claims:
  Amount: 1.50000000 BTC

Referenced transaction:
  Amount: 0.15000000 BTC

The PSBT's claimed UTXO information does not
match the referenced transaction.
```

If the required transaction is unavailable:

```text
⚠️ UTXO NOT VERIFIED

The referenced transaction is not available,
so the PSBT's UTXO claim cannot be independently
verified.
```

Once all inputs are independently verified, the transaction fee could also be reported as independently verified rather than merely calculated from PSBT claims.

## Initial scope

This does not need to fetch anything from the network.

A first version could verify `WITNESS_UTXO` directly where the relevant transaction output can be established, and verify `NON_WITNESS_UTXO` against the PSBT input's outpoint.

If a referenced previous transaction is not available, EntropyLab should report the input as **unverified** rather than treating the PSBT claim as authoritative.

Three clear states would be useful:

* ✓ **Verified** — the UTXO data matches the referenced transaction.
* ⚠️ **Unverified** — the referenced transaction/data is unavailable.
* ❌ **Mismatch** — the supplied or embedded transaction contradicts the PSBT claim.

BIP-174 defines the `NON_WITNESS_UTXO` and `WITNESS_UTXO` fields used to provide UTXO information in a PSBT:

[...]

PR #253, opened 2026-09-01
PSBT Editor: add semantic comparison engine

### LINKED ISSUE #252: PSBT Editor: Add semantic diff / comparison
## Problem

PSBTs are commonly passed between multiple participants and implementations before a transaction is finalized.

For example:

```text
PSBT A
  ↓
Coordinator
  ↓
Signer / hardware wallet
  ↓
PSBT B
  ↓
Another signer
  ↓
PSBT C
```

Bitcoin Core's PSBT documentation describes this multi-party workflow, including creator, updater, signer, and finalizer roles:

* [[BIP 174 / Bitcoin Core PSBT documentation](https://github.com/bitcoin/bitcoin/blob/master/doc/psbt.md)](https://github.com/bitcoin/bitcoin/blob/master/doc/psbt.md)
* [[Bitcoin Core multisig tutorial](https://github.com/bitcoin/bitcoin/blob/master/doc/multisig-tutorial.md)](https://github.com/bitcoin/bitcoin/blob/master/doc/multisig-tutorial.md)

Bitcoin Core also provides [`[combinepsbt](https://github.com/bitcoin/bitcoin/blob/master/src/rpc/mempool.cpp)`](https://github.com/bitcoin/bitcoin/blob/master/src/rpc/mempool.cpp) for combining PSBTs representing the same underlying transaction.

Before passing a PSBT to another participant, combining it, or signing it, it can be useful to know **what changed from the previous PSBT**.

A byte-level comparison is not sufficient because PSBT serialization order can differ while the underlying PSBT data is semantically equivalent. `bitcoinjs-lib`, for example, has tests covering PSBTs with different serialized representations that remain logically equivalent:

* [[bitcoinjs-lib PSBT tests](https://github.com/bitcoinjs/bitcoinjs-lib/blob/master/test/psbt.spec.ts)](https://github.com/bitcoinjs/bitcoinjs-lib/blob/master/test/psbt.spec.ts)

## Proposal

Add a semantic comparison/diff view to the existing PSBT Editor.

Given two PSBTs, compare their decoded contents and show meaningful differences rather than comparing the raw serialized files.

For example:

```text
PSBT COMPARISON

Transaction
✓ Version unchanged
✓ Inputs unchanged
✓ Outputs unchanged
✓ Amounts unchanged
✓ Locktime unchanged

Input #0
✓ UTXO unchanged
✓ Derivation unchanged
+ Partial signature added

Input #1
✓ UTXO unchanged
⚠ Derivation information changed

Global
+ Field added

Result:
✓ Underlying transaction unchanged
⚠ PSBT metadata/signing state changed
```

If the transaction itself changed:

```text
Outputs

Output #1
Before: 0.10000000 BTC
After:  0.09000000 BTC

Result:
❌ Underlying transaction changed
```

## Initial scope

The first version could be intentionally small:

1. Accept two PSBTs.
2. Parse both using the existing PSBT infrastructure.
3. Compare the transaction data.
4. Compare global, input, and output PSBT maps.
5. Report fields as unchanged, added, removed, or changed.
6. Compare decoded structures rather than serialized ordering.
7. Distinguish transaction changes from PSBT metadata/signing-state changes.
8. Add test vectors for semantically equivalent PSBTs with different serialization ordering.
9. Add tests for meaningful transaction, metadata, and signature changes.

No signing or network functionality would be required.

## Why this may be useful

This would provide a simple way to inspect PSBTs that have passed through another participant or implementation.

Potential examples include:

* reviewing a PSBT before signing;
* checking what another signer added;
* comparing a PSBT before and after a wallet operation;
* inspecting PSBTs before combining them;
* debugging interoperability between PSBT implementations.

The comparison itself should remain descriptive rather than attempting to assign a generic "safe/unsafe" score.

## Relationship to existing PSBT work

This appears complementary to the existing PSBT Editor and analysis work.

* The PSBT Editor answers **"what is in this PSBT?"**
* UTXO verification answers **"is this UTXO information correct?"**
* Security analysis answers **"what can we establish about this PSBT?"**
* A semantic diff would answer **"what changed between these two PSBTs?"**

The comparison primitive could also provide a foundation for more advanced PSBT analysis later without requiring additional signing or wallet functionality.

Hero-Gamer:
## Summary

Implements the semantic PSBT comparison proposed in #252, wired into the PSBT Editor as a "Compare with another PSBT" section.

The comparison operates on decoded PSBT structure — both sides are compared as fresh rust-bitcoin decodes of accepted PSBTs — so semantically equivalent PSBTs with different map ordering are not incorrectly reported as different, and mid-edit field text in the editor cannot fake a difference. It reports differences as added, removed, or changed, and separates transaction, signing-state, and metadata changes.

## What's included

- Compare transaction-level fields and global, input, and output PSBT maps; detect added, removed, and changed fields; distinguish transaction, signing, and metadata changes; ignore serialization/map ordering
- Editor UI: paste a second PSBT, get a grouped report (Transaction / per-input / per-output / global maps) with git-style +/−/~ marks, a summary banner, and a fee line when the PSBT-claimed fee moved. The report is descriptive only; it does not judge whether a change is safe
- The PSBT tools' per-tab Address network selects are removed; both read the header network picker directly, and the editor re-decodes addresses live when the picker changes
- Tests: real-fixture round-trips (a reordered, byte-different build compares equal; an output edit flags the transaction only; a removed partial signature is signing-only), report escaping against untrusted input, and browser coverage of the full compare flow in Firefox and Chrome

## Notable fix over the original engine

The unsigned transaction rides in the global map as key 00 (PSBT_GLOBAL_UNSIGNED_TX); diffing it there double-reported every transaction change as a metadata change. That pair is now excluded from the map diff (the transaction section covers it), with a regression test.

## Known v1 limitations (intentional)

- Inputs/outputs are matched **by index**, not by prevout
- Map values are compared as raw hex primitives as produced by `psbtInspectDoc`
- The engine only **reports** differences; it does not decide whether a change is safe

Verified with `npm run build`, `npm run test:ci`, `npm run test:browser` (Firefox + Chrome), and `npm run verify`.

Closes #252


Hero-Gamer:
Quick note on process: the original implementation in this PR (the psbt-diff.js engine and its tests) was drafted with ChatGPT's help. The latest commit, cbd3047 ("Fix scope/index on transaction field diffs"), was a fix I made with Claude's help after review — it corrects a bug where compareTransactionField hardcoded scope: "transaction" and dropped the index field for input/output changes, which also caused the "transaction changes are reported separately" test to fail against the original implementation.

itme-brain:
The branch is behind rock and has merge conflicts, but they're small:

1. Drop entropylab.html and SHA256SUMS.txt from the diff — those are generated artifacts we don't take in PRs (they get rebuilt by CI).
2. Rebase onto current rock — the only real conflict is package.json, where you just need to re-add test/psbt-diff.test.mjs to the test:ci script on top of the new line.

Please resolve those and we'll merge.

Hero-Gamer:
Thanks! @itme-brain 

Message from my clanker:
I've rebased the branch onto the current `rock` branch, removed `entropylab.html` and `SHA256SUMS.txt` from the PR, and resolved the `package.json` conflict while keeping `test/psbt-diff.test.mjs` in `test:ci`.

The branch should now be ready for review. Please let me know if you spot anything else that needs adjustment.

portlandhodl:
Maintainer update: I pushed `2c46520` onto this branch (fast-forward from `6ed254e`) and updated the PR description to match — it previously said "engine + tests only, no UI", which is no longer true.

What the new commit does:

1. **Engine fix:** the global unsigned-transaction pair (key `00`) is excluded from the map diff — it is the transaction section's own bytes, so counting it double-reported every transaction change as a "metadata" change on real `psbtInspectDoc` documents. Regression test included (real fixture round-trip: an output-amount edit must flag `transactionChanged` only).
2. **UI wired:** the "Compare with another PSBT" section now lives in the PSBT Editor, rendered with the editor's existing table/tone/expandable patterns. Both sides are compared as fresh rust-bitcoin decodes, so mid-edit field text cannot fake a difference and an unbuildable state refuses with an error. Report is descriptive (no safe/unsafe verdict), per the issue.
3. **Network selects removed from both PSBT tabs:** the inspector and editor read the header network picker's choice directly; the editor re-decodes addresses live when it changes.
4. **CI wiring fixed:** the earlier `test:ci` edit had dropped `test/msig-descriptor-import.test.mjs` when adding `test/psbt-diff.test.mjs`; both are listed now (the ci-completeness guard catches this).

Verified locally: `npm run build`, `npm run test:ci` (918/918), `npm run test:browser` (Firefox + Chrome), `npm run verify`.

Hero-Gamer:
Optech #420 + #421 — PSBT parsing fixes backing this issue:

- Core #35665 / #36025 / #35516 (from #420): fixes merging two global xpub records grouped by key origin vs xpub, tapscript grouped by leaf script vs control block, and `joinpsbts` dropping global xpub/metadata — creating invalid PSBT that `decodepsbt` rejects. This is exactly class of bugs #191 guards against by verifying `non_witness_utxo txid` and canonical scriptPubKey, not trusting `WITNESS_UTXO`.

- Core #36032 (from #421): makes output parsing linear not quadratic (was rescanning internal list per key). `createrawtransaction, createpsbt, sendmany` parsing 10k outputs: 1.8s → 0.5s. Relevant for offline inspection of large PSBTs in EntropyLab WASM.

Citing for context: https://bitcoinops.org/en/newsletters/2026/08/28/ and https://bitcoinops.org/en/newsletters/2026/09/04/

PR #255, opened 2026-09-01
Add QR code popups for external reference links when offline

doggyballs:
## Summary

EntropyLab is designed for air-gapped computers, so clicking an external educational link does nothing useful — there is no network. This PR adds a build-time QR system that intercepts clicks on external links when the page is offline and shows a pop-up QR code the user can scan with an online phone.

## How it works

- When offline (`#network-status[data-state="offline"]`), clicking any external `http(s)` link opens a modal with a QR code (rendered with the existing `uqr` library) and the full URL text below it
- When online, links behave normally (open in a new tab) — the hosted site is unaffected
- Link discovery is automatic via event delegation — any new external link gets QR treatment with no extra wiring, so every rebuild includes the system
- The modal follows the existing overlay pattern (`exp-overlay`, `disclaimer-overlay`): click outside or Escape to close
- Uses the existing `uqr` dependency (already bundled for SeedQR, descriptor QRs, address QRs)

## Files changed

- `src/js/qr-references.js` — new module: `isOfflineLink`, `referenceQrSvg`, `initQrReferences`
- `src/css/styles.css` — `.qr-ref-overlay` / `.qr-ref-card` styles
- `src/js/app.js` — import + `initQrReferences()` call in `hodlBoot()`
- `test/qr-references.test.mjs` — 6 unit tests for pure functions
- `package.json` — added test to `test:ci` script

## Testing

- `npm run build` — passes
- `npm run test:ci` — 844 tests, 0 failures
- `npm run verify` — site artifact verified

## Scope

All external `http(s)` links get the QR treatment when offline, including the GitHub header button. Relative links (e.g. `entropylab.html` download) are unaffected.


portlandhodl (approved):
Approved. Full-diff review of 8f62433: no bugs, exploits, or malicious code found.

Verified:
- Policy: no entropy generation, no network egress, no new dependencies (uqr is already used for address QRs); single-purpose, minimal diff; entropylab.html correctly not committed.
- Security: URL and link text are escaped before innerHTML; uqr SVG output carries no user-controlled text (verified empirically with an injection-payload URL); the ^https?:// gate keeps javascript: hrefs out of the popup; clipboard write is user-initiated.
- Integration: offline detection reuses the #network-status contract from network-check.js exactly; overlay mechanics mirror the existing disclaimer/expandable overlays; uqr options are valid and match the existing address-QR palette.
- Tests/build: new unit tests pass and are registered in test:ci (ci-completeness green); npm run build and the full suite pass — 970/0 after merging current rock (conflicts in package.json test:ci and the app.js boot list were resolved as simple unions; both init calls preserved).

Non-blocking follow-ups for a later PR:
1. "Copy URL" silently no-ops where the Clipboard API is unavailable (e.g. file:// in some browsers) — the main seed-copy path has an execCommand fallback, this could reuse it. Low impact: the URL is shown as selectable text.
2. uqr throws RangeError for URLs beyond QR capacity (~2331 bytes at ECC M); preventDefault has already run, so a >2.3KB href would be a dead click. A try/catch in openOverlay would harden it. Unreachable with the current in-document links.
3. The new test file's comment says initQrReferences is covered by test/browser-suite.html — it isn't. Add coverage there or drop the claim.
4. CONTRIBUTING §5 asks for a README note for user-facing changes; the offline QR-popup behaviour isn't documented there yet.

PR #263, opened 2026-09-01
Make incomplete PSBT security analysis explicit

### LINKED ISSUE #217: Make incomplete PSBT security analysis explicit
### Problem

EntropyLab performs increasingly comprehensive security analysis of PSBTs, including UTXO/amount handling, SIGHASH policy, signature/nonce analysis, output derivation, and Taproot-related checks.

For a security analysis tool, there is an important distinction between:

* **A check completed and found no issue**
* **A check found a problem**
* **A check could not be completed**

Some existing analyses already fail safely when they cannot fully evaluate the relevant data. For example, nonce analysis does not report a clean result when finalized signatures cannot be sufficiently decoded or associated with a key.

However, the limitation of an incomplete analysis is not always exposed as clearly or consistently as the positive result.

This creates a potential ambiguity for users:

> Does “no issue detected” mean the relevant data was fully analyzed, or was some of it simply not analyzable?

For an offline verification tool, I think that distinction is worth making explicit.

### Proposal

Consider adding an explicit **incomplete / partial analysis state** to the existing PSBT security analysis where an analysis cannot fully evaluate the relevant data.

For example:

```text
PSBT SECURITY ANALYSIS

✓ UTXO / input association    Verified
✓ SIGHASH policy              Verified
✓ Output derivation           Verified
⚠ Nonce analysis              Incomplete

Input #2 could not be fully analyzed because
the required signature/key information was unavailable.

Overall:
⚠ ANALYSIS INCOMPLETE

A clean security conclusion cannot be given
for the incomplete analysis above.
```

The exact UI and internal representation can be determined based on the existing architecture. The important invariant is:

> **An analysis that did not fully execute should not be indistinguishable from an analysis that completed successfully and found no issue.**

### Scope

For an initial implementation, I suggest keeping this deliberately narrow:

* Identify existing analyses that already have meaningful incomplete/unsupported cases.
* Expose those cases explicitly to the user.
* Preserve existing clean and warning behavior.
* Continue to fail closed for security-sensitive checks.
* Add regression tests for complete, warning, and incomplete cases.
* Avoid introducing a generalized security score or a new PSBT analysis subsystem.

This should complement existing PSBT work rather than duplicate it.

In particular, this is **not intended to replace or duplicate #191 (Independent PSBT UTXO Verification)**.

### Why this is relevant now

This fits EntropyLab's broader direction toward offline verification rather than merely parsing PSBT data.

The same issue also appears in the wider Bitcoin ecosystem. Bitcoin Core's PSBT tooling is intended to communicate the state of a cooperative signing workflow, and recent development has exposed cases where the information available to an analyzer does not necessarily map cleanly to whether a signer can actually proceed.

For example:

* [[Bitcoin Core PSBT documentation](https://github.com/bitcoin/bitcoin/blob/master/doc/psbt.md)](https://github.com/bitcoin/bitcoin/blob/master/doc/psbt.md) describes `analyzepsbt` as reporting the current PSBT status and next workflow step.
* [[Bitcoin Core #36035](https://github.com/bitcoin/bitcoin/issues/36035)](https://github.com/bitcoin/bitcoin/issues/36035) discusses a Taproot multisig case where `analyzepsbt` reports missing input information even though the PSBT can be signed.
* [[Bitcoin-Dev discussion on signer validation](https://groups.google.com/g/bitcoindev/c/GTIO4xDX5MU/m/yTrnHE54CgAJ)](https://groups.google.com/g/bitcoindev/c/GTIO4xDX5MU/m/yTrnHE54CgAJ) emphasizes that signers should independently validate transaction invariants rather than relying solely on the PSBT transport/coordination layer.

These are not proposed dependencies for the implementation; they provide context for why clearly communicating the limits of analysis is increasingly important.

### Acceptance criteria

* An incomplete PSBT security analysis is clearly distinguishable from a completed clean analysis.
* Existing warning behavior is preserved.
* Existing fail-closed behavior is preserved.
* The UI does not imply that an incomplete check passed.
* Regression tests cover complete, warning, and incomplete analysis paths.
* The change remains compatible with EntropyLab's offline, non-signing design.

I'm happy to take on the implementation/PR once the maintainers have had a chance to decide whether this direction and scope make sense.

J25dunn:
Fixes #217.

The PSBT inspector now starts with an explicit per-check status summary. Each listed check is marked Completed, Problem found, or Incomplete, and any partial coverage produces an overall ANALYSIS INCOMPLETE result. Existing blocking policy and repeated-nonce warnings remain visible and keep their severity.

The summary maps the inspector's current boundaries without turning them into a security score:
- previous-output amounts and fees remain unverified PSBT claims;
- output ownership needs a session key and remains bounded to accounts 0–2, 50 receive/change addresses, and supported script types;
- finalized/Taproot policy data that cannot be evaluated is incomplete;
- nonce coverage is incomplete for unreadable signatures, fewer than two comparable ECDSA signatures, missing key/digest data, unsupported scripts, and Taproot/Schnorr signatures;
- Taproot inscription scanning reports decode failures as incomplete.

“Completed” is explicitly scoped to the information available in the file. It does not claim that PSBT data is true or that the transaction is safe to sign. README and SECURITY document the same boundary.

Regression coverage includes pure complete/problem/incomplete status cases and a browser assertion that the existing BIP-174 fixture is presented as incomplete rather than clean.

Verification:
- `npm run test:ci` — 852 passed
- `npm run build`
- `npm run verify`
- Chrome/Chromium browser suite — all 64 checks passed

The installed snap Firefox could not stage its profile because its required host runtime directory is read-only in this environment. Microsoft Edge is not installed.


Hero-Gamer:
Hey @J25dunn thanks for taking on the PR. 
I ran it through Claude for a bug check it found something, perhaps you can speak to your AI to see if it's a real bug. 
(Cannot get my Grok to push the fix automatically - for some reasons its looping itself.. )

---

**Bug: overall banner can never show “ISSUES FOUND” / `psbt-bad`**

In `hodlPsbtAnalysisSummary` the aggregate status is computed as:

```js
let incomplete = checks.some((check) => check.state === "incomplete"),
    problem = checks.some((check) => check.state === "problem"),
    overall = incomplete ? "ANALYSIS INCOMPLETE" : problem ? "ISSUES FOUND" : "LISTED CHECKS COMPLETE",
    overallClass = incomplete ? "psbt-warn" : problem ? "psbt-bad" : "psbt-ok";
```

`incomplete` is evaluated first. The first entry in the `checks` array (“Previous outputs and fee”) is **unconditionally** `state: "incomplete"` (by design — fee/amount claims are never independently verified). Consequently `incomplete` is always `true` for every PSBT, the `problem` branch is dead code, and the top-level banner can never become `"ISSUES FOUND"` / `psbt-bad`.

### Concrete impact
When a real problem is detected (e.g. `reused.length` > 0 → nonce reuse that leaks a private key, or `policyProblems` from a malformed/unsafe SIGHASH), the individual row correctly shows `state: "problem"`. The summary banner, however, still reads **ANALYSIS INCOMPLETE** (orange) instead of **ISSUES FOUND** (red). This contradicts the PR’s stated goal that “Existing blocking policy and repeated-nonce warnings remain visible and keep their severity.”

Row-level logic already gets the priority right:

```js
state: reused.length ? "problem" : nonceIncomplete ? "incomplete" : "complete",
state: policyProblems ? "problem" : policyIncomplete ? "incomplete" : "complete",
```

Only the aggregate reverses the precedence.

### Why the new tests miss it
The browser-suite assertions only exercise the incomplete-but-no-problem path (`ANALYSIS INCOMPLETE`, the three “Incomplete” rows, `!includes("LISTED CHECKS COMPLETE")`). There is no case that asserts a problem state coexisting with the always-incomplete “Previous outputs and fee” check.

### Required fix
Flip the precedence so a detected problem dominates:

```js
overall = problem ? "ISSUES FOUND" : incomplete ? "ANALYSIS INCOMPLETE" : "LISTED CHECKS COMPLETE",
overallClass = problem ? "psbt-bad" : incomplete ? "psbt-warn" : "psbt-ok";
```

(Optionally keep both signals in the label, e.g. `"ISSUES FOUND — analysis also incomplete"`, but `overallClass` **must** resolve to `psbt-bad` whenever any check has `state === "problem"`.)

### Required regression test
Add a fixture (or mutate an existing one) that triggers a genuine problem (reused nonce **or** unsafe SIGHASH) while the always-incomplete “Previous outputs and fee” check is still present. Assert that the banner text is `"ISSUES FOUND"` (or the combined wording) **and** that the banner element carries the `psbt-bad` class, not `psbt-warn`.

Please keep the per-row priority (`problem` before `incomplete`) consistent with the aggregate after the change.

J25dunn:
Thanks for catching this — the report was correct. Because previous-output verification is always incomplete, the aggregate did make the problem branch unreachable.

Fixed in 2b06757: a detected problem now takes precedence and the combined state renders as a red `ISSUES FOUND — ANALYSIS ALSO INCOMPLETE` banner, while preserving the incomplete rows. I added both a focused aggregate regression and a browser regression that mutates the existing finalized-signature fixture from SIGHASH_ALL to SIGHASH_NONE and asserts the red combined banner.

Verified locally with:
- `npm run test:ci` (852 passed)
- `npm run build`
- `npm run verify`
- Chrome/Chromium browser suite (all 64 checks passed)

Issue #286, opened 2026-09-03
i18n: fully automate translations with validated PRs and reproducible builds


dplusplus1024:
## Decision

Translation-only changes must be **fully automated**. There is no human translation review, no `CODEOWNERS` gate for locale files, no manual Actions approval, and no person pressing merge. Feature and code PRs still follow the repository's normal policy; this decision applies only to the tightly scoped translation PRs created by the translation workflow.

## Problem

Translation coverage falls behind whenever English UI text changes. Requiring feature authors to update every language encourages hardcoded English, while translating during the build would make the shipped artifact non-reproducible.

## Pipeline

```text
Feature PR merges into rock
├─ normal CI builds and deploys immediately
│  └─ missing or stale locale keys display the committed English fallback
└─ post-merge translation workflow runs
   ├─ Job A: generate and validate translations (LLM secret, read-only repo access)
   └─ Job B: revalidate and open one translation-only PR per language
      └─ required CI passes → GitHub auto-merges → normal build runs again
```

Translation is a side loop that produces new commits. It is never a build step. Every artifact remains a pure function of the committed tree and can be reproduced without an LLM, API key, or network access.

## Required foundations

1. **Sanitize catalog output in the app.** Rich translations are rebuilt from a small allowlist (`code`, `strong`, `em`, and `span.mono`); attribute values and placeholders are escaped in their final context. Prefer `textContent` wherever markup is unnecessary.
2. **Make `src/locales/en.json` the single source of truth.** `npm run i18n:sync` deterministically updates inline English fallbacks, and CI rejects drift or referenced keys missing from English.
3. **Allow partial locales.** Missing and stale keys fall back to English per key without hiding the language.
4. **Track staleness independently per language.** `src/locales/.sources/<lang>.json` stores `sha256(English text)` for every translated key. A translation PR changes exactly one catalog and its matching sidecar.
5. **Validate all catalog changes in CI.** Reject empty values, malformed source hashes, changed placeholders, disallowed tags or attributes, URLs or executable schemes, bidi overrides, zero-width characters, and unexpected controls. Missing and stale keys are reported and use English fallback. Keys removed from English are reported as obsolete and made inert so the feature PR can merge; the translation workflow removes each obsolete catalog value and its sidecar hash together.

## Fully automated translation workflow

[...]

PR #287, opened 2026-09-03
i18n: rebuild catalog markup from an allowlist before it can reach the page (step 1 of #286)

dplusplus1024:
First step of #286: make the catalogs safe to fill by machine before anything fills them by machine.

## The problem

Translated text reaches the DOM through two paths that trust it: `data-i18n-html` assigns a catalog value to `innerHTML`, and most runtime markup interpolates `hodlT()` values into template strings that are assigned to `innerHTML`. A catalog value containing `` would run in the page. That is true today for a hand-written translation from any contributor; it becomes the central question once translation drafts are produced by an LLM job, as #286 proposes.

## What this does

**A rebuild-from-allowlist sanitizer, applied once at catalog load.** `src/js/i18n-sanitize.js` tokenizes a value and reconstructs it from a closed alphabet: the elements the catalogs actually use — `code`, `em`, `strong`, `span class="mono"`, and the one `a href="entropylab.html" download="entropylab.html"` in the online banner — with attributes only at exactly those values. Anything else — an unknown tag, an extra or altered attribute, a comment, a script — is emitted as escaped text, so a bad string is visible on screen and can never execute. Tag names and attributes in the output come from the allowlist table, never from the input, which is what makes this a boundary rather than a check.

Values with no tag-like sequence pass through byte for byte. That keeps `textContent` and attribute sinks exact, and keeps the four origin-path errors that mention `<0;1>` readable (a `<` followed by a digit is text to an HTML parser and stays text here).

`i18n.js` sanitizes all five catalogs where they are imported. Attribute-context template call sites use `hodlTAttr()`, `t()` re-sanitizes after placeholder interpolation, and a CI guard rejects future raw `hodlT()` calls inside quoted template attributes.

**A validator that runs in CI as a lint**, `test/i18n-catalog.test.mjs`: every translated value must be a valid rendering of its English source — same tags, same placeholders, only allowlisted markup, no `<script` / `on*=` / `javascript:` / `data:`, no bidirectional, zero-width or control characters (they pass a tag check but can make a string render differently from how it appears in source). Non-English catalogs may omit keys — missing keys are reported and fall back to English at runtime — so an English-only feature PR passes. Length outside 0.3×–3× of the English is a warning, not a failure. A source-hash sidecar is validated when present, ahead of the process PR that introduces it.

## Tests

- `test/i18n-sanitizer.test.mjs`: passthrough of plain text including the bare `<0;1>`; allowlisted markup survives exactly; every shape of executable or foreign markup (event handlers, `javascript:` hrefs, altered `download`, `<svg>`, comments, unquoted and self-closing tags, unterminated and mis-nested tags) is emitted as text; a closed-alphabet check over hostile inputs; the sanitizer is idempotent and **leaves every current value in all five catalogs unchanged**; a source guard that `i18n.js` routes every catalog through it and that `t()`'s English fallback reads the sanitized copy.
- `test/i18n-catalog.test.mjs`: the validator above, over all five catalogs.
- `test/i18n-attribute-guard.test.mjs`: parses real JavaScript template interpolations across line breaks, follows nested templates, recursively scans every `src/js/**` subdirectory, and rejects raw `hodlT()` calls in quoted attributes while allowing `hodlTAttr()` and DOM `setAttribute()` sinks.
- `test/browser-suite.html`: in Firefox, Chrome and Edge, a hostile catalog value assigned to `innerHTML` produces no `img`, no handler attribute, no link, does not execute, and keeps its allowlisted `strong`/`span.mono`; legitimate catalog markup (`msig.intro`) still renders as markup.

Every current catalog value is a fixed point of the sanitizer, so this changes nothing users see today.

## Audit of runtime interpolation

Checked where `hodlT()` receives variables derived from state or input. The values that reach markup are internal labels and prefixes; the one user-typed value (`keys.renameAria` with the key name) is set through `setAttribute` and an input's `value`, not `innerHTML`. Runtime values interpolated directly into templates outside `hodlT()` already go through `hodlEscapeHtml` where they are user-controlled; that path is unchanged by this PR and remains a separate concern from the catalogs.

## Out of scope, by design

The English source-of-truth, sync, partial-locale fallback, and source-hash sidecars are in #289. Remaining static-shell wiring and the fully automated translator workflow stay in later rollout steps. This security boundary lands first.

## Verification

- `npm run test:ci` — 983 tests: 976 passed and 7 intentionally skipped.
- `npm run build` and `npm run verify` — passed.


MrHodlX:
Superseded by #310 (sanitizer + partial-locale model — the foundations of parts 1–3, rebuilt on the post-#307 architecture) and #311 (the post-merge translation automation, completing #286). The #307 i18n rewrite made this line un-rebasable; suggest closing in favor of the new stack.

portlandhodl:
Superseded by #312, which lands the sanitizer against the content-keyed catalog architecture (no en.json). Your work is credited there: test/i18n-attribute-guard.test.mjs lands byte-identical from this PR, the sanitizer suite largely derives from yours, and the three browser-level hostile-catalog assertions are ported forward in the consolidation commit (Co-authored-by: D++). Thanks for the step 1–3 line — it defined the shape of what shipped.

PR #288, opened 2026-09-03
Add a Vanity tab: deterministic address grinder (WASM + Web Workers)

portlandhodl:
## What

Adds a **Vanity workspace tab** (between Keys and BIP-85) that grinds vanity addresses from a deterministic counter — a calculator, never an entropy source:

- **Counter as passphrase.** Counter *i* maps to a fixed-width base-62 odometer string over `a-zA-Z0-9` ("aaa…", "aab…"); that string is the brain-wallet passphrase (`privkey = SHA-256(passphrase)`), and the derived address is checked against the user's prefix. Same counter, same address — every result replays by hand.
- **Entropy import.** A one-click **Import from the Keys tab** salts the grind with the session: the active key's passphrase verbatim when one is entered (a found passphrase is that passphrase followed by the counter characters), else a SHA-256 digest of the session's non-empty entropy inputs (only the digest crosses into workers and results). A salt can also be typed directly. With no salt, the UI states plainly that the counter space is public.
- **All four single-signature script types** — legacy P2PKH, nested SegWit P2SH-P2WPKH, native SegWit P2WPKH, Taproot P2TR — selected on the tab; the prefix field is live-filtered to base58 or bech32 characters and shows a 1-in-*N* work estimate. Address matching is mainnet; a found passphrase derives on any network.
- **`vanity-wasm/`** — new pinned Rust crate (libsecp256k1 0.8.0 via `secp256k1-sys =0.14.0` with the full precomputed EC tables, RustCrypto `sha2 =0.10.9` / `ripemd =0.1.3`, committed `Cargo.lock`, `rust-toolchain.toml`), committed artifact `src/js/vanity-wasm-b64.js` via the existing `scripts/build-wasm.mjs` flow. The module derives all four address types itself; private keys never leave the loop — only a matching candidate's counter, passphrase, and address payload cross into JS.
- **Threading** — one Web Worker per CPU core, spawned from an inline Blob source (`src/js/vanity-worker.js` exports the source string; CSP gains `worker-src blob:`). No `SharedArrayBuffer`, so it works from `file://` without COOP/COEP and the output stays a single self-contained `entropylab.html`. 4,096-candidate chunks keep Stop responsive; the counter space splits into disjoint contiguous buckets (no overlap, no gaps), and the start counter auto-advances after each run.
- **Safe output** — matches render as a table (counter / passphrase / address). Passphrases are masked behind the standard reveal toggle ("air-gap only"), copy buttons read live match state rather than DOM attributes, and found passphrases plus the imported salt join the existing pagehide/bfcache secret clearing. The grinder stops when the tab is left. Results point the user at Keys → Private key → Brain-wallet text for the WIF and every script type.

## Relation to #161 and #213

This implements the shared core of the two earlier grinder proposals as one tab: #161's engine and threading model (dedicated crate, worker pool, buckets, session salt) with an explicit import action and per-tab script selection, without #213's CSPRNG lab mode (section 1 of CONTRIBUTING: new exceptions belong in an issue first) and without its GPU/WGSL path, keeping the audit surface small.

## Tests

- `test/vanity-wasm.test.mjs` (22 tests): artifact integrity/no-host-paths; pinned counter → passphrase → address vectors for all four script types re-derived independently with `@noble/curves` + Node crypto + `@scure/btc-signer`; session-salt re-keying; the full 1-character odometer space; bucket-split equivalence; prefix filtering; argument rejection (incl. the 256-byte salt cap); bucket partition disjointness/coverage; the **shipped worker source** executed under `node:worker_threads` (init/grind/progress/done + cooperative stop); the `VanityGrinder` pool aggregating buckets.
- `test/ui-defaults.test.mjs`: the tab is registered and localized (5 locales), card/intro ship hidden in both templates, nothing auto-runs, the reveal/copy conventions hold, the `worker-src blob:` CSP pin, and no `Math.random`/`getRandomValues` anywhere in the vanity code paths. Existing tab-strip order assertions now cover seven tools.
- `test/secret-clear.test.mjs`: the wipe drops the salt field, matches, reveal state, and cancels the worker pool.
- `test/browser-suite.html`: tab keyboard navigation updated for seven tools, plus a new end-to-end test that grinds prefix "14" over the 1-character space in the real page and finds the pinned address `14dD6ygPi5WXdwwBTt1FBZK3aD8uDem1FY` — passing in headless Firefox and Chrome (exercises the blob worker under the CSP and the in-worker WASM instantiation).

## Commands run

```sh
npm run build:wasm   # all three crates; vanity artifact built from the pinned sources
npm run build        # entropylab.html, 4,621,988 bytes
npm run verify       # site artifact verified
npm test             # 928 pass — the only failures are the untracked WIP test/psbt-network.test.mjs, which fails identically on clean rock
```

CI wiring: `test:ci` lists the new suite; the `build-wasm` job rebuilds all three crates and runs `test/vanity-wasm.test.mjs` against the fresh build; the `artifact` job commits `src/js/vanity-wasm-b64.js` back alongside the other artifacts; the Dockerfile pre-fetches the vanity crate's dependencies.


portlandhodl:
Review follow-up (3a05e2b): entropy now comes in through the shared Key Station chip picker (same clickable LifeHash chips as BIP-85 / Silent Payments / MultiSig), and the salt contract is verbatim-only — the grind extends the user's passphrase with the counter characters, never rehashes session entropy into a digest prefix. The digest fallback and `vanitySessionSalt` are removed. Also pinned in tests: grinding is WASM-only (no JS curve/hash loop, no CPU fallback — fails closed if workers are blocked).

portlandhodl:
Fix (448d526): the import chips could render zero entries and hide the picker — the filter required a passphrase to already exist. The picker now lists every Key Station tab (like the BIP-85/SP pickers list theirs), and a passphraseless key explains itself when picked. This also exposed a latent boot hazard worth noting for reviewers: a boot-time picker fill hit `typeof hodlLifeHash?.fromFingerprint`, which throws `ReferenceError` on an undeclared identifier when the WASM-ready promise settles before the later parser-inserted lifehash.js tag (deterministic sanity-failure in headless Chromium). `hodlFillKeyTabLifehash` now leads with the plain `typeof === "undefined"` guard, matching the footer's documented pattern; nothing vanity-related fills at boot.

PR #290, opened 2026-09-03
MyFirstBanana

unclejim7:


dplusplus1024:
I like the idea of documenting a dedicated offline setup, but I don't agree with the threat model in the hardware section.

In particular:

> a machine's history and provenance don't add security here, only the fact that it will never touch a network again

I don't think that's true. An air gap reduces the attack surface, but it doesn't make an already-compromised machine trustworthy. Malware could already be present, and compromise can exist outside the disk in firmware/UEFI or other components. Fully wiping the disk doesn't address those cases.

Similarly, disabling Wi-Fi/Bluetooth in the OS or BIOS/UEFI isn't equivalent to knowing that the machine has no functioning radio hardware. If we're describing a higher-assurance setup, physically removing the Wi-Fi/Bluetooth hardware is meaningfully different.

I'd rather frame this around levels of assurance:

* Prefer a machine with known history/provenance.
* Wipe it and install a verified OS from verified media.
* Physically remove Wi-Fi/Bluetooth hardware where practical.
* Keep the machine permanently offline after setup.
* Treat software/BIOS radio disabling as additional defense, not proof of an air gap.

A brand-new laptop isn't automatically trustworthy either, of course, but I don't think we should tell users that provenance is irrelevant. For a machine handling bitcoin private key material, it's one of the things that can contribute to confidence in the system.

I also think it's fair to say explicitly that securely configuring a laptop for this purpose isn't for everyone, and it isn't necessarily easy. There are a lot of details involved in getting meaningful assurance that the machine is actually offline and trustworthy.

dplusplus1024:
* **Speakers:** Encode secrets into audible or ultrasonic sound that can be received by a nearby microphone or device.
* **Keyboards:** Individual keys can produce distinguishable acoustic signatures, potentially allowing typed secrets to be reconstructed.
* **Fans:** Malware can vary fan speed to encode data into fan noise or physical vibrations.
* **Hard drives:** HDD actuator movement can be manipulated to produce encoded acoustic signals.
* **Power supplies:** PSU components can be manipulated to produce encoded acoustic emissions.
* **Power lines:** CPU power consumption can be modulated so encoded data appears as measurable fluctuations in electrical wiring.
* **USB/SATA/Ethernet cables:** Cables can unintentionally act as antennas and emit controllable electromagnetic signals.
* **LEDs:** Keyboard, disk, NIC, or other status LEDs can blink encoded data that a camera or optical sensor can capture.
* **CPU electromagnetic emissions:** CPU activity can be manipulated to transmit information through RF or magnetic fields.
* **Heat:** CPU temperature or workload can be modulated and detected by another nearby machine as a very slow covert channel.

**The larger point:** "Offline" eliminates normal network exfiltration, but it does not inherently eliminate side-channel exfiltration from a compromised computer.

dplusplus1024:
Then there's also attack vectors that don't assume any compromise, even with WiFi and Bluetooth off:

### Passive side-channel leakage — no compromised computer required

- **Keyboard acoustics:** Keystrokes produce different sounds that can reveal which keys are being pressed.
- **Electromagnetic emissions:** CPUs, displays, cables, buses, and other electronics naturally emit electromagnetic signals that can reveal information being processed.
- **Power analysis:** Variations in a computer's power consumption can correlate with cryptographic operations and potentially reveal secret information.
- **Acoustic emissions:** Electronic components can naturally emit sounds correlated with computation, including coils and power-supply components.
- **Vibrations:** Mechanical activity from keyboards, fans, drives, and other components can leak information through surfaces.
- **Display leakage:** Electromagnetic or optical emissions from a display can potentially reveal what is being shown on screen.
- **Thermal leakage:** Computation changes component temperatures and heat patterns, which can expose information about system activity.

portlandhodl:
PR scope needs to be split up. 

We have securing your device and 501c in the same pull.

unclejim7:
HA yea im an idiot. Will fix when i can

PR #299, opened 2026-09-03
feat: from mnemonic to deck of cards and vice versa

### LINKED ISSUE #300: feat: store a mnemonic using one or more deck of poker cards
## Problem
There is no way to back up or recover a mnemonic (BIP39 seed phrase) using poker cards.

dani69654:
# Reversible BIP39 Card Backup

## Summary

This PR adds a `Card Backup` workspace to EntropyLab. It converts a valid
English BIP39 mnemonic into a deterministic order of ordinary playing cards and
recovers the mnemonic from that order. The feature is an encoding of existing
user-supplied secret material; it does not generate entropy.

## Motivation

The existing Playing cards method is intentionally one-way: it hashes a card
transcript into BIP39 entropy. That is useful for creating a seed from a
shuffle, but an existing mnemonic cannot be turned back into the original card
sequence. A physical card order is a useful plausible-looking offline backup,
provided the method is public, deterministic, and honest about its limits.

## Encoding

The implementation uses lexicographic Lehmer-code ranking and unranking of
permutations. No secret-dependent hash or random generator is used.

- 12, 15, 18, and 21 words use one standard 52-card deck.
- 24 words use a complete first deck and the first 6 cards of a second deck.
- The remaining 46 cards of the second deck are the canonical cards not in the
  six-card prefix, in canonical order.
- The word count must be recorded with a one-deck backup because a card order
  does not contain a separate length field.

A 52-card permutation contains `log2(52!) = 225.58` bits. This is enough for
the first four BIP39 entropy sizes, but not for 256-bit entropy. The second
deck prefix supplies the missing capacity for a 24-word mnemonic. Recovery
rejects malformed decks, duplicate cards, non-canonical second-deck tails, and
out-of-range ranks.

## Passphrases

An arbitrary BIP39 passphrase cannot be recovered from a finite fixed-size card
order. The UI accepts a passphrase and displays a short verification marker,
but the passphrase is deliberately not encoded in the cards. Users must keep it
as a separate secret. This avoids implying that the card backup alone restores
the wallet when a passphrase is in use.

## Security notes

- The card order is a wallet backup. Anyone who knows the method and possesses
  the cards can recover the mnemonic.
- The physical deck must be kept private and its order must not be photographed
  or exposed during preparation.
- This feature is not a BIP39 extension or an officially approved card format.
  BIP39 specifies entropy, checksum, and word encoding, not a physical entropy
  source or reversible card backup format.
- The existing hash-based Playing cards modes are unchanged.
- The app remains self-contained, air-gapped, and free of runtime network
  requests.

## Tests

Added tests cover:

- round trips for all five BIP39 word counts;
- uniqueness and full-deck shape;
- the zero-entropy 24-word vector;
- rejection of a one-deck 24-word recovery;
- passphrase marker behavior without exposing the passphrase.

Run with:

```sh
npm run build && npm test
```

https://github.com/user-attachments/assets/81f234b9-776a-4b85-87a7-c32b9ac3bdb4

closes https://github.com/OogaBoogaX/entropylab/issues/300


dani69654:
Hey @portlandhodl , would you consider this one to be useful to users? If you believe so I will be glad to rebase to master and improve it if needed. Else please feel free to close it.

w-s-bitcoin:
Thank you for building this out and for clearly documenting the permutation approach. It is an inventive experiment, but we’re closing it because it adds a new wallet-backup format and workspace during the v1 feature freeze, and the current design has a recovery-safety issue we are not comfortable shipping.

The deck order directly encodes entropy without an integrity check that binds the complete order. In review, encoding the standard `abandon … about` mnemonic and swapping only the final two cards recovered a different checksum-valid mnemonic (`abandon … actual`) while the passphrase marker still matched. That means a plausible physical handling or transcription mistake can silently restore a valid but entirely different wallet. The four-byte unsalted passphrase marker also acts as a fast offline guessing oracle, is linkable across backups, and is too short to serve as strong integrity protection.

This is also a proprietary recovery format without a portable versioned specification stored alongside the cards. The branch now conflicts with `rock` and predates the current source/build and translation rules, so bringing it forward would require a redesign rather than a routine rebase. We do not want you to spend time rebasing this implementation in its present form. Thank you again for the contribution and for checking with us before doing more work on it.

PR #301, opened 2026-09-03
Draft: BIP-47 reusable payment codes, calculator only (lands after feature freeze)

MrHodlX:
**Draft. This lands after feature freeze. Do not mark ready, do not merge.**

BIP-47 reusable payment codes, as a calculator tab. No chain access of any kind.

Four commits, rebased onto `rock` at `bae51d8` (the `psbt-analysis-status` merge). The only rebase conflict was the `test:ci` list in `package.json`, where both sides added a suite; both are present. `npm test` is 997/997 on the rebased tree.

## Research summary

Read before any code was written: `README.md`, `CONTRIBUTING.md`, `SECURITY.md`, `AGENTS.md`, `llms.txt`, the whole of `src/js/`, `src/index.html`, `src/locales/`, and the suites that encode the contract (`bip85.test.mjs`, `bip352.test.mjs`, `secret-clear.test.mjs`, `wipe-wasm.test.mjs`, `ui-defaults.test.mjs`, `ci-completeness.test.mjs`, `i18n.test.mjs`, `journal.test.mjs`).

What that established, and what this change follows:

- The lab is a calculator, never a generator. `bip47.js` contains no CSPRNG call and no default that ends up in secret material. Every output is a deterministic function of what the user supplied.
- No network egress. `bip47.js` has no `fetch`, no `XMLHttpRequest`, no `WebSocket`, no storage. A notification transaction is pasted in as data.
- One secp implementation. This adds **no new WASM and no new dependency**. ECDH is the existing `secp256k1.Point.multiply`, the notification blinding factor is the existing `hashes.js` `hmacSha512`, payment codes ride the existing rust-bitcoin `base58ck` facade, and addresses come from the existing rust-bitcoin `Address` facade.
- A new tab is wired in seven places (static markup in `src/index.html`, the `hodlWorkspaceTabs` array, the panel template in `app.js`, the `hodlShowWorkspace` toggles, a Station bench chip, `hodlRefreshStationKeyPickers`, and the Journal tool map plus session snapshot). All seven are done here, matching Silent Payments exactly.
- Scheme crypto lives in its own DOM-free module so it is testable from Node. `bip47.js` follows `bip352.js`; `app.js` holds only UI glue.
- Locale catalogs must stay key-identical or the locale is dropped from the picker. Both new keys were added to all five files.

The BIP itself was read, not recalled: `bip-0047.mediawiki` and the vector gist it links (SamouraiDev/6aad669604c5930864bd).

Two things the vectors settled that the prose does not:

1. The BIP writes the notification blinding HMAC with its arguments in one order for the sender and the other for the recipient. The published mask settles it: **key = the 36-byte outpoint, data = the secret point's x value**. There is a test that pins exactly this.
2. The gist's `S0..S9` column is the **secret point's x value**, not `SHA256(S_x)`. Both are surfaced so a reader can follow either column against the table.

## What was added

- `src/js/bip47.js` — payment code serialization and Base58Check (version byte `0x47`), the `m/47'/coin_type'/identity'` path, notification key and address, ECDH send and receive derivation, notification blinding and unblinding, OP_RETURN payload extraction, and address verification against a derived window.
- A **BIP-47** workspace tab between Silent Payments and PSBT, with a BIP-47 Station chip and the Key Station chip picker, four modes:
  - **My code** — payment code, notification address, payment code public key and chain code, the 80-byte payload. The notification private key stays hidden behind an explicit reveal.
  - **Send** — the P2PKH addresses this pair produces, plus an optional collapsed section that blinds the payment code into the 80-byte OP_RETURN payload for a notification transaction.
  - **Notification** — recovers the sender's payment code from pasted notification data and derives the receive addresses that follow, with the spending keys behind a reveal.
  - **Verify** — checks a pasted address against a derived window in either direction.
- `test/bip47.test.mjs` (26 tests) and `test/fixtures/bip47-vectors.json`.
- Session clearing, wipe coverage, Journal session-state section, `workspace.bip47` / `workspace.bip47Short` in all five locales, and `README.md` / `SECURITY.md` / `llms.txt`.

## What was explicitly not added

- No chain scanning, no node, no Electrum, no indexer, no lookup of any kind.
- No transaction building, no input selection, no signing, no broadcasting. The Send tab produces an OP_RETURN payload and says outright that the transaction must be built and signed elsewhere.
- No CSPRNG, no lab opt-in mode, and no new exception to the entropy rule.
- No GPU, WebGPU, SharedArrayBuffer, or COOP/COEP.
- No MuSig2, FROST, or SeedXOR. None were needed.
- No Bitmessage notification (features bit 0). The features byte is parsed and reported; setting it is out of scope.
- No version 2 notification flow. The version byte is accepted and reported, but the bloom-filterable multisig change output is a wallet concern, not a calculator's.
- No change to the single-HTML build contract, no new dependency, no new WASM, no version bump.
- `entropylab.html` is not touched. It is generated, `.gitignore`d, and rebuilt and committed by CI after merge.

## Watch-only and the refusal to invent secrets

An xpub gives the public half and nothing more: the payment code and the notification address. Every ECDH step is refused with a specific reason, because the shared secret genuinely needs a private key from one side:

- A root **xpub** is refused for payment code derivation at all, because `m/47'` is hardened and cannot be reached from it. The message says so and points at pasting the payment code node itself.
- A depth-3 **xpub** derives the payment code and notification address, then refuses send, notification recovery, receive addresses, and private keys.
- `deriveReceiveAddresses` refuses a public-only node outright rather than returning something derived from nothing.

## Wipe behavior

- The notification private key is the only secret a derive hands back. `hodlBip47WipeResult()` zeroes it; `hodlBip47WipeKeys()` also calls `wipePrivateData()` on the session root itself, not on a getter copy.
- Inside the module, `derivePaymentCodeKeys` wipes the payment code node and the notification node it derived, and `deriveReceiveAddresses` wipes each child's private getter copy and each tweaked receive key after rendering. `HDKey.derive` already wipes intermediate path nodes.
- `hodlBip47WithAccountNode` derives a throwaway private node for a root session and wipes it in a `finally`; a pasted account node is the session's own and is left for the session wipe.
- `hodlInitSecretFieldAutoClear` calls `hodlBip47WipeMem()` and blanks `#bip47-key`, `#bip47-pass`, `#bip47-designated-key` (the private key of a notification input), `#bip47-out` (which renders private material when revealed), `#bip47-error`, `#bip47-session`, `#bip47-recipient`, `#bip47-payload`, `#bip47-verify-code`, and `#bip47-verify-address` on `pagehide` and on a bfcache restore.
- `test/secret-clear.test.mjs` and `test/wipe-wasm.test.mjs` gained assertions for all of the above.

## Journal

The session snapshot gains a BIP-47 section carrying the fingerprint, the payment code, the notification address, and a watch-only marker. Payment codes and notification addresses are published by design, so they ride along regardless of the include-private setting. **No seed, no xprv, and no notification or receive private key ever enters the snapshot**, with or without include-private. Tool switches, calculates, mode changes, copies, and the session clear are logged as `bip47` events.

## How to test offline

```sh
npm ci
npm run build      # src/ -> entropylab.html
npm test           # or: npm run test:ci
npm run verify
```

Then open `entropylab.html` from the filesystem with the network off and walk the vectors by hand:

1. **BIP-47 tab -> My code.** Paste Alice's mnemonic `response seminar brave tip suit recall often sound stick owner lottery motion`, leave the passphrase blank, identity 0, mainnet. Derive. Expect payment code `PM8TJTLJbPRGxSbc8EJi42Wrr6QbNSaSSVJ5Y3E4pbCYiTHUskHg13935Ubb7q8tx9GVbh2UuRnBc3WSyJHhUrw8KhprKnn9eDznYGieTzFcwQRya4GA` and notification address `1JDdmqFLhpzcUwPeinhJbUPw4Co3aWLyzW`.
2. **Send.** Paste Bob's payment code `PM8TJS2JxQ5ztXUpBBRnpTbcUXbUHy2T1abfrb3KkAAtMEGNbey4oumH7Hc578WgQJhPjBxteQ5GHHToTYHE3A1w6p7tU6KSoFmWBVbFGjKPisZDbP97`, first index 0, count 10. Expect `141fi7TY3h936vRUKh1qfUZr8rSBuYbVBK` at index 0 through `1GxfdfP286uE24qLZ9YRP3EWk2urqXgC4s` at index 9, and the secret point x values `f5bb8470…` through `fe36c27c…`.
3. **Send -> blind.** Designated key `Kx983SRhAZpAhj7Aac1wUXMJ6XZeyJKqCxJJ49dxEbYCT4a1ozRD`, txid `9c6000d597c5008f7bfc2618aed5e4a6ae57677aab95078aae708e1cab11f486`, vout 1. Expect the payload to start `010002063e4eb95e62791b06c50e1a3a942e1ecaaa9afbbeb324d16ae6821e0916…`.
4. **Notification.** Reload with Bob's mnemonic `reward upper indicate eight swift arch injury crystal super wrestle already dentist`, paste the blinded payload from step 3, designated pubkey `0272d83d8a1fa323feab1c085157a0791b46eba34afb8bfbfaeb3a3fcc3f2c9ad8`, same txid and vout. Expect Alice's payment code back and the same ten addresses.
5. **Verify.** Paste `1KQvRShk6NqPfpr4Ehd53XUhpemBXtJPTL` with the counterparty code and direction set. Expect a match at index 4.
6. **Watch-only.** Paste an xpub. Expect the payment code and notification address, and a refusal with a stated reason on every other mode.
7. **Wipe.** Reveal private material, then hide the tab or close it. Every field and the output region come back empty.

## Vector sources

- `bitcoin/bips` `bip-0047.mediawiki` (Status: Deployed, finalized 2021-02-15).
- The vector gist the BIP links: <https://gist.github.com/SamouraiDev/6aad669604c5930864bd>, transcribed into `test/fixtures/bip47-vectors.json` with the source URL recorded in the file.

`@scure/bip39` and `@scure/bip32` are used in the test only as independent oracles for the seeds and the BIP32 walk, matching how `bip85.test.mjs` and `bip352.test.mjs` already use them. They stay devDependencies.

## It does not scan the chain

Stated plainly because it is the thing most likely to be assumed otherwise. This tab has no node, no indexer, no Electrum client, and no network call. It cannot tell you whether an address was ever paid, whether a notification transaction was ever sent, or what your next unused index is. A "no match" in Verify only means the address was not in the window derived on screen. Receiving payments still requires a node or a wallet elsewhere. That is said on the tab itself, in `README.md`, in `SECURITY.md`, and in `llms.txt`.

## One repo contradiction, surfaced rather than forced

`test/ui-defaults.test.mjs` pinned the workspace tab order as contiguous literal substrings of `hodlWorkspaceTabs`. Two of those assertions together pinned all seven existing tools as one unbroken run with Journal last, leaving no gap for an eighth tab except ahead of Keys.

Adding BIP-47 in its natural position, after Silent Payments, required **extending two assertions** (and the `["bip85", "sp", "msig", "calc", "vanity"]` tool-intro membership list in the Vanity suite) to include the new entry. The assertions still pin the complete order; they were not loosened, and nothing was skipped or deleted. Flagging it here rather than doing it quietly, because "do not weaken a test" is the rule and this is the closest thing to it in the change.

`test/browser-suite.html` needed the same kind of update, and it is worth calling out because it caught a genuine break: it asserted `tabs.length === 7` and named all seven accessible names in order, so an eighth tab failed the live keyboard-navigation check in all three engines. The count is now 8, `BIP-47` is in the names array in its position, and the two hard-coded "last tab" indexes became `tabs[tabs.length - 1]` so they keep meaning what they said. Verified by running the suite on clean `rock` (passes), then on this branch before the fix (fails in Firefox, Chrome and Edge), then after (passes).

If the maintainers would rather the tab sat somewhere else, or would rather those assertions were left exactly as they were, say so and I will move it.


MrHodlX:
Closing as superseded by #443.

Same BIP-47 v1 calculator, parked until after 0.2.0. #443 keeps it on the Silent Payments tab instead of adding a new workspace tab.

Branch `feat/bip47-calculator` stays on the fork for now in case any of the watch-only/xpub-refusal or journal snapshot notes are still useful there.

PR #292, opened 2026-09-03
i18n: wire the remaining static shell (step 3 of #286)

dplusplus1024:
## Summary

- Wires the current static-shell UI copy through `src/locales/en.json`.
- Makes the wiring gate structural: CI re-parses the current source every run, with no fixed element count, approved-element list, or committed baseline.
- Handles the complete key lifecycle dynamically: additions fall back to English, edits make old translations stale, and removals make old locale entries inert until automation deletes them.
- New wired elements pass immediately, removed elements simply disappear, and any new hardcoded visible text fails CI until it has an English key.
- Keeps protocol/build literals immutable and uses placeholders for technical values inside translated rich text.
- Excludes runtime-owned wallet/session/journal outputs so changing locale cannot overwrite live results.
- Guards dynamic Key, multisig, and Journal tab deletion against stale active rename editors.
- Enforces exact placeholder bindings, safe element ownership, and literal invariants in CI.

## Security and reproducibility

This PR does not call a model or modify any translated catalog. Missing or stale translations continue to fall back to English; removed keys are made inert. The build remains a pure function of the committed tree.

This branch is stacked on #289 at `b8da30c`. Merge #287 first, then #289; after those land, this PR reduces to its four Step 3 commits.

## Testing

- `npm run test:ci` — 1,004 tests: 997 passed, 7 expected skips; build and artifact verification passed.
- `npm run test:browser` after a fresh build — 73/73 Firefox, 73/73 Chrome, and 73/73 Edge.
- `npm run i18n:check` — zero unwired static-shell entries; each existing non-English locale has 323 missing, zero stale, and zero obsolete keys, so English fallback is used until Step 4 translates them.

PR #289, opened 2026-09-03
i18n: make English the source of truth (step 2 of #286)

dplusplus1024:
## Summary

- Makes `src/locales/en.json` the authoritative source for committed UI fallback text.
- Adds deterministic `npm run i18n:sync` and CI drift checks for the HTML and JavaScript templates.
- Adds per-language SHA-256 source sidecars and English fallback for missing or stale translation keys.
- Keeps incomplete languages selectable without displaying translations known to be stale.
- Adds `title` and `alt` attribute wiring plus a guard against new hard-coded root-template UI text.
- Documents the deterministic translation lifecycle, including that stale-key filtering is injected into the built artifact and that browser tests require a fresh build.

## Security and reproducibility

No model, API secret, network call, or build-time translation is introduced. The build remains a pure function of committed source files. Fully automated translation belongs to Step 4 of #286.

This branch is stacked on #287 at `75340ad`. Merge #287 first; once that commit is in `rock`, this PR reduces to the Step 2 commit only.

Step 2 retains a temporary migration baseline; #292 removes it and replaces it with a structural scan of the current source on every CI run, so UI elements may be added or removed without maintaining a fixed inventory.

## Testing

- `npm run test:ci` — 997 tests: 990 passed and 7 intentionally skipped; build and artifact verification passed.
- Browser integration is exercised from a fresh build on the final stacked tip in #292 across Firefox, Chrome, and Edge.

PR #310, opened 2026-09-04
i18n: sanitize catalogs at load and let partial locales ship

MrHodlX:
Refs #286 (foundations 1–3). Supersedes the sanitizer work in #287 and the per-key-fallback work in #289, rebased onto the content-keyed catalogs from #307 — there is no `en.json`; the English source text is the catalog key, so the old `en.json`-based steps could not be rebased mechanically.

## What lands here

- **`src/js/i18n-sanitize.js` (new)** — every catalog value is rebuilt from a markup allowlist at load, in two views: an HTML view (allowlisted formatting only, entity-escaped text) and a text view (markup dropped, safe for `textContent`/`setAttribute`). Tags map to complete fixed attribute *forms*: attributes render from the table, never from input, so even a misrouted `tHtml()` value cannot end a quoted host attribute. The anchor forms pin exactly the twelve links the committed English sources carry — a translation can never introduce or retarget a link.
- **`src/js/i18n.js`** — `t()`/`tAttr()` read the text view, `tHtml()` the HTML view; placeholder values are escaped per view. **The completeness gate is gone**: every locale is always selectable and missing strings fall back to English per string. Partial locales are the normal state between a UI change and the next automated translation run — which is what lets the post-merge automation ship.
- **`src/js/app.js`** — ~130 call sites routed to the sink-correct helper. `hodlPublicFieldHtml`/`hodlPrivateFieldHtml` now translate and escape their labels internally, so raw translator calls stay out of template interpolations. A handful of previously hardcoded labels (wallet field labels, journal field labels) become real translatable sources; they will report as missing translations until the automation fills them.
- **`scripts/i18n-validate.mjs` (new)** — the merge gate for catalog content: placeholder parity with the English source, exact allowlist forms, balanced nesting, anchors limited to the source's own links, canonical entities only, no bidi overrides / zero-width / control characters, no executable schemes, no empty values (untranslated means the key is absent).
- **`scripts/i18n-sync.mjs`** — missing and dead entries are now **reported, never failed**, so feature PRs no longer touch locale catalogs. Invalid catalog content and source markup outside the sanitizer table (a new link or formatting form) still fail CI — that is the tripwire that keeps the allowlist complete.
- **`test/i18n-attribute-guard.test.mjs` (new)** — statically rejects translator calls in unquoted or wrong-view template positions across `src/js`.
- **`AGENTS.md`** — the contributor flow now says: don't touch catalogs.

## Testing

- `npm ci && npm run build && npm test` → **1053/1053 pass** (includes the browser suites and the new sanitizer/attribute-guard/validator suites).
- `npm run verify` → artifact verified.
- All four committed catalogs pass the validator; every translated anchor now survives sanitization (8 of 10 anchor-bearing values were being escaped under the original two-form allowlist — fixed by pinning the full committed link set).

## Notes for review

- The catalogs themselves are deliberately untouched; the only locale-visible change is richer rendering of the already-committed links plus a few newly translatable labels.
- #287 / #289 / #292 can be closed once this lands — their content is covered here against the new architecture.
- The post-merge automation (issue #286 part 4) stacks on top of this PR.


portlandhodl:
Superseded by #312, which includes this PR's content as commit 1 (4ffa3b6, your authorship preserved) rebased onto current rock, plus the fixes the rebase exposed (BIP-321/BIP-353 anchors in the sanitizer allowlist, sidecar reseed) and the restored browser-level sanitizer tests.

PR #316, opened 2026-09-04
Journal: log a bad-nonce verdict from PSBT inspect

MrHodlX:
## Summary
When PSBT / raw-tx inspect finds a reused or possible ECDSA nonce, write that verdict into the Journal session log. Surface the same short state in the session snapshot.

Does **not** store r values, signatures, keys, or hex. Detail is counts only (`kind · comparable · reuse · possible · unreadable`).

## Behavior
- `inspect-nonce-reuse` — same key, different digest (blocking reuse)
- `inspect-nonce-possible` — same r, digest could not be rebuilt
- `inspect-nonce-incomplete` — some ECDSA sigs were unreadable
- Session state PSBT block adds a one-line nonce verdict when a inspect has run
- Wipe of the PSBT session clears the verdict
- Clean / single-sig inspect does not spam the log

## Security
- Inspect-only. No signing. No CSPRNG. No network. No browser storage.
- Log and snapshot never include r, DER, or pubkeys.

## Note for reviewers
`app.js` is too large to rewrite through this connector in one shot. The complete source diff is in `patches/journal-bad-nonce.patch` (app.js, journal.js, journal.test.mjs, ui-defaults.test.mjs). Apply that patch onto this branch before merge.

## Validation
- `node --test test/journal.test.mjs test/psbt-nonce-reuse.test.mjs` passed on the patched tree
- Not merged. Review only.

### DIFF
--- a/patches/journal-bad-nonce.patch
+++ b/patches/journal-bad-nonce.patch
@@ -1,0 +1,137 @@
+diff --git a/src/js/app.js b/src/js/app.js
+index 8d7ea5f..f342095 100644
+--- a/src/js/app.js
++++ b/src/js/app.js
+@@ -8224,6 +8224,33 @@ function hodlCompareNonces(rValues) {
+   }
+ }
+ 
++// Last inspect-only nonce verdict for the Journal session snapshot.
++// Never stores r values, signatures, or keys — counts and a short state only.
++var hodlPsbtNonceVerdict = "";
++function hodlJournalNonceVerdict(kind, reused, possible, uninspected, comparable) {
++  let detail = `${kind} · ${comparable} comparable · ${reused.length} reuse · ${possible.length} possible · ${uninspected} unreadable`;
++  if (reused.length) {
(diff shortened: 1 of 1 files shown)


MrHodlX:
Superseded by #341, which applies this feature as a real src/ diff (this branch committed it as an inert patches/*.patch file that nothing consumes) and fixes the review findings F1-F5: clean-verdict gating now matches the on-screen incompleteness predicate, the nonce verdict is cleared on failed re-inspect, snapshot lines are gated on a loaded inspector and kind-qualified, and the journal hook runs after the warning banners inside try/catch. Suggest closing this one.

w-s-bitcoin:
Closing as superseded by #341. That PR applies this work as an actual source diff instead of an inert `patches/*.patch` file and incorporates the review fixes for clean-verdict gating, stale verdict clearing, snapshot gating, kind-qualified output, and failure-safe journal logging. Thanks for preserving the original investigation and implementation context here.

PR #319, opened 2026-09-05
Validate BIP352 input private keys

Rob1Ham:
## Summary
- derive a public point from every supplied BIP352 input private key
- require it to match the public key extracted from that input
- compare Taproot inputs by x-only key so parity-equivalent scalars remain valid

## Security impact
Before this change, a valid but unrelated private scalar produced a plausible Silent Payment output. The receiver derives from the transaction input public keys and could not detect that output, creating a funds-loss path if it was funded.

Replacing the first private key in published vector 0 with scalar 1 changed the output from `3e9fce73...` to `41ba134c...`; the recipient found zero matching outputs.

## Tests
- `node --test test/bip352.test.mjs`
- `npm run test:ci` (1,094 passed)

### DIFF
--- a/src/js/bip352.js
+++ b/src/js/bip352.js
@@ -364,14 +364,20 @@ const normalizeVin = (vin) => ({
 export function eligibleInputKeys(vins) {
   const pubkeys = [];
   const privkeys = [];
-  for (const raw of vins) {
+  for (const [index, raw] of vins.entries()) {
     const vin = normalizeVin(raw);
     const extracted = extractInputPubKey(vin);
     if (!extracted) continue;
     pubkeys.push(extracted);
     if (vin.private_key) {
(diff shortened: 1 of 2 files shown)


portlandhodl:
Merged, with a companion commit (4415537) that the new check exposed as necessary: the session-derived send flow injected the raw internal key for P2TR inputs, which your validation correctly rejects — BIP-352 requires "the private key corresponding to the taproot output key (i.e. the tweaked private key)". Before this pair of changes, a taproot session send silently produced an output the recipient could never detect (verified: send produced 1 output, scan found 0). hodlSpDeriveVinKeys now injects the BIP-341-tweaked scalar for P2TR inputs, and a send->scan roundtrip regression test proves detectability. The official vectors keep passing: every eligible taproot input in them has an output key equal to the raw key's x-only form (the two mismatches in vector 19 are NUMS-point inputs, which are skipped as ineligible before the check).

Issue #328, opened 2026-09-05
Do not label malformed PSBT fields signed or finalized


Rob1Ham:
## Severity
Low security-state presentation risk

## Summary
The PSBT visualization derives signing status only from pair names. A malformed partial signature or final script field retains a known name even when typed decoding fails, so the UI can display `1 signature` or green `finalized` for invalid data.

## Affected code
- `psbt-wasm/src/lib.rs:589-598` retains pair identity alongside decode errors
- `psbt-wasm/src/lib.rs:648-653` exposes named pairs
- `src/js/psbt-viz.js:75-82` counts names without requiring successful decode
- `test/psbt-viz.test.mjs:90-100` currently uses name-only fields

## Impact
A malicious PSBT can overstate signing progress and induce attempted finalization or export of unusable data.

## Acceptance criteria
- count signing fields only after successful typed decoding
- label field presence separately from cryptographic validity
- for future signing, maintain distinct present, structurally valid, cryptographically valid, policy-complete, and finalized states
- add malformed ECDSA, Taproot, scriptSig, and witness regression cases

PR #330, opened 2026-09-05
Drop PSBT signing material when a transaction edit orphans it

### LINKED ISSUE #325: Revalidate PSBT signatures after transaction edits
## Severity
Medium now; required before in-browser signing

## Summary
The PSBT editor regenerates the unsigned transaction after edits but copies all input-map pairs unchanged. Partial signatures, Taproot signatures, final scriptSigs, and final witnesses remain attached even when committed transaction fields change. The final gate only deserializes the PSBT and does not verify signatures against the rebuilt transaction.

## Affected code
- `src/js/psbt-editor.js:249-260` copies all input maps unchanged
- `src/js/psbt-editor.js:803-885` permits transaction and structural edits
- `psbt-wasm/src/lib.rs:868-911` passes input fields through
- `psbt-wasm/src/lib.rs:917-920` checks only `Psbt::deserialize`
- `test/psbt-wasm.test.mjs:132-146` mutates a finalized fixture and accepts parse success

## Reproduction
Changing an output amount in the finalized BIP174 fixture preserves `PSBT_IN_FINAL_SCRIPTSIG`, and `rustBitcoinError` remains null.

## Acceptance criteria
- block export when transaction edits leave unvalidated signing/finalization fields
- either remove affected fields or cryptographically revalidate them with sighash-aware logic
- cover ECDSA and Taproot sighash variants in tests
- keep field presence separate from signature validity

itme-brain:
## The bug

The PSBT editor changed the unsigned transaction — an output amount, a prevout, a sequence, locktime or version — without touching the input maps, and `psbt_build` regenerated `PSBT_GLOBAL_UNSIGNED_TX` from the edited fields while passing every key-value pair through verbatim. `Psbt::deserialize` still accepted the result, so **the file went out looking signed**: partial signatures and final scripts over a transaction nobody authorized, under a green "Rebuilt PSBT is accepted by rust-bitcoin" verdict.

Reproduced empirically against `test/fixtures/psbt/p2wpkh-1in-2out.b64` (which carries `PSBT_IN_PARTIAL_SIG`) and BIP-174 valid vector 2 (which carries `PSBT_IN_FINAL_SCRIPTSIG`): after an output-value edit, a sequence change, an output deletion or a full prevout repoint, the signature pairs came back byte-identical with `rustBitcoinError: null`.

Closes #325.

## The fix

The editor now keeps the transaction section's **normalized key** from each successful decode and drops the pairs that assert signing material when it moves:

- `PSBT_IN_PARTIAL_SIG`, `PSBT_IN_FINAL_SCRIPTSIG`, `PSBT_IN_FINAL_SCRIPTWITNESS`, `PSBT_IN_TAP_KEY_SIG`, `PSBT_IN_TAP_SCRIPT_SIG`

What went is named in a notice above the pair tables (`input 0 PSBT_IN_PARTIAL_SIG (pubkey 0279be667e…f81798)`) — no full signature bytes echoed to the DOM.

Three details that matter:

- **`PSBT_IN_SIGHASH_TYPE` is deliberately kept**, and so are a Taproot input's `PSBT_IN_TAP_INTERNAL_KEY` / `PSBT_IN_TAP_BIP32_DERIVATION`. These declare *who may* sign rather than attesting that anyone did, so a transaction edit leaves them alone for re-signing.
- **The drop happens on a draft and is adopted only when rust-bitcoin accepts the build**, matching the existing `mutate()` semantics: a state that does not build keeps the working document, signatures and all, exactly as it was.
- **Field normalization** (string vs number, upper-case hex) means a keystroke that did not change a value is not counted as a transaction edit, so an unedited round-trip still keeps every pair.

Detection lives in JS, not the WASM gate, because that is where a baseline exists — `psbt_build` has no record of the transaction a signature was made over. No Rust change, no `build:wasm` needed.

## Against #325's acceptance criteria

| Criterion | How it lands |
| --- | --- |
| block export when edits leave unvalidated signing fields | Satisfied by construction: the orphaned pairs are removed from the document that is built, so no unvalidated signing field survives into base64, hex, `.psbt` or QR. The exported file is honestly unsigned rather than exporting a blocked artifact. |
| remove affected fields **or** revalidate with sighash-aware logic | Removal, the first branch. Cryptographic revalidation in the editor was rejected: it needs the full script/interpreter context per spend type and would duplicate the analyzer's work for no extra safety over removal. |
| cover ECDSA and Taproot sighash variants in tests | Both: `PSBT_IN_PARTIAL_SIG` (ECDSA), `PSBT_IN_FINAL_SCRIPTSIG` (finalized P2PKH, BIP-174 vector 2) and `PSBT_IN_TAP_KEY_SIG` (`test/fixtures/psbt/p2tr-taproot.b64`). |
| keep field presence separate from signature validity | Untouched by design — `psbt-viz.js` `signingStatus()` is not modified here. Note this criterion overlaps **#328**, which is about name-only status labelling of *malformed* fields; that is a genuinely separate fix and this PR neither claims nor regresses it. |

## Tests

- **`test/psbt-editor-stale-signing.test.mjs`** (new, 11 tests, wired into `test:ci`): each committed field drops; the ECDSA, finalized and Taproot cases; the no-edit round-trip keeps every pair; the sighash-type / internal-key exemptions; normalization; purity and the identical-object no-op.
- **`test/browser-suite.html`**: asserts the drop in input 0's *own* table — the "Add pair" select lists every pair type, so a whole-output `textContent` probe cannot see a removal (this bit me once). The compare test now expects `Signing state changed`, which is what a transaction edit that dropped its final script really is; it previously pinned the buggy behaviour.

```
docker compose run --rm dev npm run build && npm test
# tests 1108 / pass 1107 / fail 0 / skipped 1   (Firefox + Chrome included)
```

`entropylab.html` is the rebuilt artifact, consistent with how CI keeps `rock` in sync.


portlandhodl:
This is a UI / UX question. IMO this would be good default behavior but we should retain a checkbox, 'enable insane editing' this box when checked would enable editing that would be considered consensus invalid. 

Really what this is adding consensus checks to the PSBT editor. Which we should do, but at the same time we would need to add other checks along the way as well. 

 - Validation of witness UTXOs (Fees)
 - Validation of non-witness UTXOs (Fees)
 
 Closing: keep insane editing, but default to a consensus editor.
 
 Pros
  - More user friendly
  
 Cons
   - Defeats the purpose of a the raw editor
   - Would limit use in fork situations possible (Sighash types)

portlandhodl:
Adding: This is not a bug and 100% intentional, but non the less a bad UX.

portlandhodl:
Closing as superseded. The bug this fixes (issue #325) is already fixed on `rock` in a more complete form — #375 plus follow-ups (`6943aff`, `0942410`, `4a3cf3a`): the landed signing anchor also covers UTXO-claim edits (BIP-143 commits to the claimed amount; this PR's tx-section-only anchor misses those — verified on this branch: bumping a witness-UTXO amount by one sat leaves the now-invalid partial signature in place) and the MuSig2 session pairs (0x1b/0x1c), and it rolls the anchor back on rejected structural edits.

Worth crediting: this PR's pure draft-then-adopt drop avoids the anchor-rollback bug rock's first version needed two follow-ups for — the design here was the cleaner one. The branch is also 212 commits behind `rock` and conflicts in four of its five files.

The follow-on layer this class of fix points at — checking every edit against Bitcoin's consensus rules and BIP-174's signer checks, listing the problems in the UI, with an "Insane editing" opt-out — is now up as #444.

PR #344, opened 2026-09-05
Fix wallet.dat hardened branch caching and signing

Rob1Ham:
## Summary
- parse the path between each descriptor extended key and wildcard
- cache the descriptor root key itself when a hardened branch already moved the xpub to branch level
- derive and store the matching branch xprv for private wallet exports
- add independent hardened BIP32 reference tests for watch-only and signing records

## Security and recovery impact
With the Harden branch option, the watch-only descriptor root is already the branch xpub. The exporter instead cached child 0 or 1 beneath that key, so Bitcoin Core watched `xpubBranch/0/i` instead of `xpubBranch/i`, disagreeing with EntropyLab addresses and potentially missing funds.

Private exports recorded the account key while Core looked up the branch root pubkey, so the generated wallet could not sign.

## Tests
- `node --test test/wallet-export.test.mjs`
- `npm run test:ci` (1,094 passed)

### DIFF
--- a/src/js/app.js
+++ b/src/js/app.js
@@ -1441,6 +1441,9 @@ function hodlWalletDatDeps() {
       body.set(node.publicKey, 41);
       return body;
     },
+    // Hardened-branch private key records are derived from the account xprv.
+    deriveExtendedPrivateChild: (extendedKeyText, index) =>
+      hodlHDKey.fromExtendedKey(hodlReversionExtendedKey(extendedKeyText, hodlExtendedKeyVersions.mainnet.x.prv)).deriveChild(index).privateExtendedKey,
     publicKeyForPrivate: (secret) => hodlSecp256k1.getPublicKey(secret, true)
   };
 }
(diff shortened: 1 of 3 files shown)


portlandhodl:
Merged with conflict resolution against the harness refactor from #372: the PR's inline reference crypto was re-based onto the shared helpers in test/wallet-export-harness.mjs (hdMasterFromSeed/hdDeriveHardened/serializeExtendedKey), and the new deriveExtendedPrivateChild dep moved into the harness so both suites share it. The fuzz verifier had to learn the same rule this PR teaches the exporter: it previously derived the expected cache body from the branch SLOT (internal ? 1 : 0), which both sides got wrong for descriptors whose text says otherwise (the corpus's forced digit-h vectors use /0/* for the change descriptor). Core derives the cache from the descriptor text, so the verifier now parses the tail independently. Also verified the Core semantics claim against BIP32PubkeyProvider (empty path after the root key caches the root key itself; signing looks up the stored descriptor's root pubkey).

PR #349, opened 2026-09-05
PSBT: an impossible (negative) claimed fee escalates the summary

Rob1Ham:
## Summary

When a PSBT's claimed output amounts exceed its claimed input amounts (a negative "fee" — an inconsistency in the PSBT's *own* data), the inspector showed a red inline paragraph (`Inconsistent claimed amounts: outputs exceed claimed inputs by …`), but the **analysis summary** kept the "Previous outputs and fee" check hardcoded at `state: "incomplete"`, so the banner stayed yellow ("ANALYSIS INCOMPLETE") instead of red ("ISSUES FOUND").

An impossible fee is a *found problem*, not missing coverage — the banner should reflect that.

## Proposed fix

```diff
     if (fee >= 0n) html.push("…Unverified fee…");
-    else html.push("<p class='psbt-bad'>…outputs exceed claimed inputs…");
+    else {
+      feeInconsistent = 1;
+      html.push("<p class='psbt-bad'>…outputs exceed claimed inputs…");
+    }
```

```diff
     {
       label: "Previous outputs and fee",
-      state: "incomplete",
+      state: feeInconsistent ? "problem" : "incomplete",
       detail: feeInconsistent
         ? "The claimed output amounts exceed the claimed input amounts, so the PSBT's own claims are inconsistent; …"
         : …
     },
```

## Tests

`test/psbt-analysis-status.test.mjs` gains a structural assertion that the render path tracks `feeInconsistent` and escalates the check to `"problem"`.

## Caveat (as requested)

Audit-produced; the escalation is structurally asserted but not driven end-to-end against a crafted negative-fee PSBT fixture. Full Docker suite green.

Commands run: `npm ci && npm run build && npm test` (in `docker compose run --rm dev`).

### DIFF
--- a/src/js/app.js
+++ b/src/js/app.js
@@ -9339,7 +9339,8 @@ function hodlRenderPsbt(psbt) {
     uninspected = 0,
     policyProblems = 0,
     policyIncomplete = 0,
-    unsupportedNonceChecks = 0;
+    unsupportedNonceChecks = 0,
+    feeInconsistent = 0;
   let inscriptionReport = { inputs: [], envelopes: [] }, inscriptionScanIncomplete = false;
   try {
     inscriptionReport = inspectPsbtInscriptions(psbt);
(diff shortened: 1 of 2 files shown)


portlandhodl:
Merged with a contextual conflict resolution: #372 reworked the surrounding fee text ("previous-output claims", the conflicted-inputs branch), so the escalation was re-applied onto the new wording. The behavior is as proposed — a negative claimed fee sets feeInconsistent and the "Previous outputs and fee" check escalates to a problem state.

Issue #331, opened 2026-09-05
Derive Silent Payment sender input keys from the loaded session


Rob1Ham:
## Priority
High before treating Silent Payment sending as a real-money workflow

## Summary
The SP Station loads a BIP39 seed or root xprv into session state, but sender output construction does not use that root to derive transaction input keys. Instead, every eligible vin is expected to carry a raw `private_key` inside pasted JSON.

PR #319 validates that each pasted scalar matches the extracted input public key, which prevents one funds-loss case, but the sender still does not follow the intended session-owned key model.

## Affected code
- `src/shell.html:548-550` asks for per-input private keys in vin JSON
- `src/js/app.js:8850-8937` loads and derives SP receive/scan session keys
- `src/js/app.js:8938-8944` parses vin JSON without origin metadata
- `src/js/app.js:9040-9044` passes pasted vins directly to sender derivation
- `src/js/bip352.js:364-406` consumes each `private_key`

## Risk
Raw key copying is easy to mismatch, bypasses the loaded session ownership model, and requires exposing every input scalar in a large JSON blob. It also provides no trusted derivation trail from the user-supplied entropy.

## Acceptance criteria
- define vin origin/path metadata sufficient to derive each session-owned input key
- derive private keys from the loaded root rather than pasted scalar fields
- verify each derived key against the prevout script and extracted input key
- reject origins outside the loaded root fingerprint or selected network
- wipe per-input derived nodes/scalars after output construction
- retain published vector support in a clearly separated test-vector/import mode if needed

Issue #322, opened 2026-09-05
Validate Bitcoin transaction sanity before PSBT export


Rob1Ham:
## Severity
Medium; required before signing support

## Summary
The editor treats successful rust-bitcoin PSBT deserialization as its authoritative gate. Deserialization validates BIP174 structure, not Bitcoin transaction sanity or consensus constraints.

The editor accepts unsigned transactions with no outputs, duplicate input outpoints, and outputs above `MAX_MONEY`.

## Affected code
- `psbt-wasm/src/lib.rs:763-817` constructs transactions without CheckTransaction-style validation
- `psbt-wasm/src/lib.rs:804` permits values through `u64::MAX`
- `psbt-wasm/src/lib.rs:917-920` gates only on PSBT deserialization
- `src/js/psbt-editor.js:580-583` reports rust-bitcoin acceptance

## Independent reproduction
EntropyLab returned `rustBitcoinError: null`; Bitcoin Core v29.2 rejected equivalent transactions with:

```text
bad-txns-vout-empty
bad-txns-inputs-duplicate
bad-txns-vout-toolarge
```

## Acceptance criteria
- separate PSBT parse validity from Bitcoin transaction validity in the UI
- enforce nonempty inputs/outputs, unique prevouts, per-output and aggregate `MAX_MONEY`, and serialized size/weight bounds before export
- add Bitcoin Core-compatible regression vectors

Issue #364, opened 2026-09-05
UR crypto-psbt fragment reassembly silently overwrites duplicate sequence numbers


Rob1Ham:
## Summary

`hodlUrDecodePsbt` (`src/js/psbt-ur.js`) reassembles sequential `ur:crypto-psbt` fragments (BCR-2020-005, seq-len form) into one CBOR payload. After validating type, fountain markers, count agreement, and the per-fragment Bytewords CRC32, it fills a fixed slot array by sequence number. A fragment whose sequence number was already stored simply **overwrites** the slot — there is no duplicate-sequence rejection.

The function therefore accepts input in which position `i` carries bytes from two different source sequences and emits whichever came last. A single-part UR is unaffected (`count === 1` requires exactly one piece), and a missing slot is still rejected, so the only silent case is **duplicate-but-complete coverage**.

## Evidence

`src/js/psbt-ur.js` lines 170–174 — slot fill:

```js
slots[part.seq - 1] = part.payload;   // line 173
```

Overwrites a previously stored fragment with the same `seq` without checking whether the slot is already filled or whether the payloads differ.

## Exploit scenario

An honest user reassembles fragments from two different PSBTs that happen to share a fragment count (e.g., a stale animated QR loop and a fresh one, or a copy that merged two sequences). Example: count 3 with fragments seqs 1,2,3,1 where the two seq-1 payloads differ — slot 1 keeps the second seq-1 fragment and the reassembly mixes PSBT A's frame 1 with PSBT B's frames 2–3. The mixed payload either fails downstream parsing (loud) or, when it still parses as CBOR+PSBT, decodes to a transaction that matches neither original — and the inspector/editor then faithfully displays that transaction.

## Impact

The PSBT shown for inspection can be a **mixture of two different PSBTs** rather than the one the user intended to inspect. Any signing decision taken on the basis of that content would rest on bytes that neither sender produced. Because the tool is inspection-only and the decoded content is displayed, the realistic impact is limited to misinformed review; no key material is derivable from the defect.

## Preconditions

The user pastes or scans a multi-part `ur:crypto-psbt` in which at least two fragments carry the same sequence number with different payloads, and every slot `1..count` is filled.

## Suggested fix

Track seen sequence numbers while filling slots and throw on a duplicate, e.g.:

```js
if (slots[part.seq - 1]) throw new Error("Duplicate UR fragment " + part.seq + ".");
```

[...]

Issue #358, opened 2026-09-05
PSBT v2 helper is dead code in production (v2 rejected; wire in or remove)


Rob1Ham:
**Reported via security audit. Informational.**

PSBT v2 (BIP-370) is not supported — `version != 0` is rejected outright (psbt-wasm/src/lib.rs:292-294), and `psbt-v2.js` (`hodlTxFromPsbtV2`) is imported only by `test/psbt-v2.test.mjs`, i.e. dead code in production. The rejection is fail-closed and correct, so this is a feature gap rather than a bug — noting it so the dead module is either wired in (with the same canonical-encoding and duplicate scrutiny as the v0 path) or removed to avoid it drifting unreviewed.

If wired in: its `PSBT_IN_PREVIOUS_TXID` is taken as raw wire-order bytes (matches Core's uint256 convention), but it currently gets no canonical-encoding or duplicate-key scrutiny beyond the shared helpers.

PR #373, opened 2026-09-05
Reject null prevouts in the PSBT transaction sanity gate

portlandhodl:
## Problem

Follow-up to #322/#361. `tx_sanity_error` (`psbt-wasm/src/lib.rs`) skips CheckTransaction's coinbase branch on the assumption that an unsigned PSBT transaction can never look like one. The editor's free-text txid/vout fields break that assumption: an input hand-edited to txid `00…00` / vout `0xffffffff` (a null prevout) builds, inspects clean, and exports under "unsigned transaction passes consensus sanity checks".

Bitcoin Core's `CheckTransaction` rejects such a transaction on both branches:

- sole input ⇒ `IsCoinBase()`, and the PSBT-forced empty scriptSig (build requires it) fails the 2..100-byte coinbase rule → `bad-cb-length`;
- one of several inputs ⇒ `bad-txns-prevout-null`.

## Change

- `psbt-wasm/src/lib.rs`: `tx_sanity_error` now rejects any null prevout with Core's `bad-txns-prevout-null` reason (the doc comment explains the `bad-cb-length` sole-input shape). The check sits in Core's order, before the duplicate-prevout loop, so it gates both the v0 and v2 build paths and the inspect report (`txSanityError`), which the editor already renders as "consensus-invalid … — export is blocked".
- `src/js/psbt-wasm-b64.js`: regenerated with `npm run build:wasm` (pinned Rust 1.95.0); the other crates' artifacts were rebuilt by the script and left out of the commit.
- `test/psbt-wasm.test.mjs`: build-side throws for the sole-input and multi-input null-prevout shapes; inspect-side, a null prevout spliced into the BIP-174 valid vector reports `txSanityError: "bad-txns-prevout-null"` with `rustBitcoinError: null` (structural vs consensus validity stay separate facts).

## Ran

`npm run build:wasm`, `cargo test --locked` (psbt-wasm), `npm run build`, `npm test` — all green (1152 pass, 1 engine-skip).


MrHodlX (approved):
SHIP.

Null-prevout check is in psbt-wasm/src/lib.rs using OutPoint::is_null, before the duplicate-prevout loop, on both v0 and v2 build paths. Inspect reports txSanityError bad-txns-prevout-null with rustBitcoinError null. Editor already blocks export on consensus failure.

Tests cover sole-input, multi-input, and a null prevout spliced into the BIP-174 vector.

Does not close #331 #328 #340 #358 #364.

Issue #350, opened 2026-09-05
PSBT inspector: fee computed from witness-UTXO claims only, no cross-check vs non-witness UTXO


Rob1Ham:
**Reported via security audit. Low severity (disclosed in UI, but asymmetric).**

The PSBT inspector builds `inputSum` **only** from `PSBT_IN_WITNESS_UTXO` claims (`src/js/app.js` ~line 9362-9366); `PSBT_IN_NON_WITNESS_UTXO` is never read anywhere in `app.js`. An input carrying **both** a non-witness UTXO (the real prev tx, whose outpoint amount is verifiable against the prev tx) and a witness UTXO claiming a different amount is summed at the witness claim with no cross-check. A deflated claim shows a small "reasonable" fee over a much larger real one; an inflated claim shows an alarming one.

The sibling editor's Rust helper **does** cross-check: `claimed_prevout_amount` (psbt-wasm/src/lib.rs:659-677) requires `outpoint.txid == prev.compute_txid()` and takes the amount from the real prevout for non-witness UTXOs. The inspector and editor are asymmetric in a security-relevant way.

**Existing mitigations:** honest labeling ("BTC claimed", "Unverified fee", permanent `incomplete` state, disclaimer line), and a partial tripwire — for already-signed SIGHASH_ALL segwit inputs, a lied amount breaks the BIP143 digest reconstruction and surfaces as a signature-verification warning.

**Residual risk:** for *unsigned* PSBTs (the common review case) there is no tripwire; the user sees a fee number derived from an unverifiable claim.

**Suggested fix:** when both utxo fields are present, prefer the verified non-witness amount, and flag disagreement between the two as a problem — matching the editor's behavior.

Related: the Rust side's `known_in_sats`/`out_sum` accumulate in u64 without overflow checks (psbt-wasm/src/lib.rs:703-710) and the editor's fee picks the first-in-file-order utxo pair (attacker-controlled); both display-only, but worth `checked_add` and a witness-preference respectively.

PR #374, opened 2026-09-05
Match non-witness UTXOs by txid, not wtxid, when the prevtx carries a witness

portlandhodl:
## Problem

Follow-up to #350. `hodlNonWitUtxo` (src/js/app.js) hashes the raw `PSBT_IN_NON_WITNESS_UTXO` bytes with double-SHA256 and requires the digest to equal the input's previous-output txid. When the embedded previous transaction is witness-serialized (BIP-144 marker/flag + witnesses), that digest is the transaction's **wtxid**, not its txid — so the checkable claim false-mismatches with "A non-witness UTXO's transaction does not match the input's previous output.", the field claims nothing, and without a witness-UTXO claim the fee goes unknown.

Witness-inclusive serialization is legal BIP-174 ("network serialization format") and common from non-Core wallets. The WASM inspector verifies the same field with `compute_txid()` (psbt-wasm/src/lib.rs), which strips the witness — so the two inspectors disagreed on the same file: the WASM side resolved the claim and computed the fee, the JS side reported a problem.

## Change

- src/js/app.js: when the embedded prevtx parses as segwit, hash `serializeTx(prev)` (the witness-stripped legacy serialization the txid actually commits to) instead of the raw field bytes. Legacy-serialized fields keep the exact old path, byte for byte.
- test/psbt-nonwitness-utxo.test.mjs: new regression test — a witness-serialized prevtx verifies against its txid (amount and script resolve), and the same bytes referenced by their wtxid still throw (the lie the check exists to catch). Verified the test fails without the fix. The harness now injects `serializeTx` alongside the other `hodlNonWitUtxo` dependencies.

## Ran

`npm run build`, `npm test` — green (1153 pass, 1 engine-skip; the two Firefox journal subtests flake under full-suite load on this machine and pass in isolation and on repeat full runs, unchanged by this diff).

### DIFF
--- a/src/js/app.js
+++ b/src/js/app.js
@@ -8039,7 +8039,13 @@ function hodlNonWitUtxo(entries, input) {
   let entry = hodlFind(entries, 0).find((item) => item.keydata.length === 0);
   if (!entry) return null;
   let prev = parseRawTx(entry.val);
-  let txid = hodlSha256(hodlSha256(entry.val));
+  // The txid commits to the legacy (witness-stripped) serialization: a
+  // witness-carrying prevtx — legal BIP-174 and common from non-Core wallets —
+  // must be stripped before hashing, otherwise this computes the wtxid and
+  // every such field is a false "does not match" (issue #350). The WASM
+  // inspector's compute_txid() strips unconditionally; stripping only the
+  // segwit case here keeps legacy bytes hashed exactly as before.
+  let txid = hodlSha256(hodlSha256(prev.segwit ? serializeTx(prev) : entry.val));
   if (!hodlEq(txid, input.txid)) throw new Error("A non-witness UTXO's transaction does not match the input's previous output.");
   let output = prev.outputs[input.vout];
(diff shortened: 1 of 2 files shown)


MrHodlX (approved):
SHIP.

At 7948fa9, witness-serialized prevtx bytes are stripped before hashing, so the honest txid matches and a wtxid outpoint still throws. Legacy transactions hash the original bytes unchanged. Verified JS/WASM agreement at 42,000 sats input and 2,000 sats fee; conflicting claims still void the fee, with WASM behavior independent of pair order. This closes #350's identified remaining serialization gap. No remaining raw-witness-as-txid path in the inspected code.

src/js/app.js:8038-8052 witness stripping, unchanged legacy branch, txid mismatch rejection.
src/js/tx.js:142-169 serialization excludes witness.
src/js/app.js:9543-9551, 9699-9704 agreeing claims count once; conflicts suppress fees.
test/psbt-nonwitness-utxo.test.mjs:96-111 honest segwit and wtxid-lie regression.

Does not expand scope beyond #350.

PR #375, opened 2026-09-05
Drop stale signatures on UTXO-claim edits, and cover MuSig2 session pairs

portlandhodl:
## Problem

Follow-up to #325. The editor drops signing pairs only when the unsigned transaction changes (`JSON.stringify(doc.tx) !== pristineTx`, psbt-editor.js). But pair values are editable too, and signatures commit to more than the transaction:

- BIP-143 commits to the spent output's **amount** and the scriptCode derived from its scriptPubKey;
- BIP-341 commits to **every** input's amount and scriptPubKey.

Editing `PSBT_IN_WITNESS_UTXO` (0x01) — or the embedded prevtx of `PSBT_IN_NON_WITNESS_UTXO` (0x00) — in the pair table left 0x02/0x13/0x14 in place, and the rebuilt file "looked signed" with invalid material. Additionally, the BIP-327 MuSig2 pairs were missing from `SIGNING_KEY_TYPES`, so even a transaction edit kept them: public nonces (0x1b) and partial signatures (0x1c) are bound to the message like any signature. (Participant pubkeys, 0x1a, are message-independent and stay.)

## Change

- New exported `signingAnchor(doc)`: the unsigned transaction **plus each input map's UTXO declarations** (0x00/0x01, key+value only — decoded/name fields are presentation). `rebuild()` compares and re-anchors on this instead of `doc.tx` alone, so a UTXO-claim edit trips `dropSigningPairs` exactly like a transaction edit. The anchor is keyed off inspected bytes, so mid-edit decorative drift cannot fake or hide a change.
- `SIGNING_KEY_TYPES` gains `1b` and `1c`, with the comment naming the BIP-327 semantics.

## Tests (test/psbt-editor-tx.test.mjs)

- MuSig2: 0x1b/0x1c drop, 0x1a stays.
- Anchor: identical across re-inspection; moves on a witness-UTXO value edit and on a transaction edit; unmoved by a sighash-hint (0x03) edit.
- The rebuild-wiring assertions now pin `signingAnchor(doc)` on both the compare and re-anchor sides.

## Ran

`npm run build`, `npm test` — green (1154 pass, 1 engine-skip).


MrHodlX (changes requested):
BLOCK.

Normal witness-UTXO amount/scriptPubKey, non-witness prevtx, and transaction edits trigger removal of 02/07/08/13/14/1b/1c; 03 edits do not, and 1a stays. However, there is an exportable stale-signature bypass: add a duplicate 01 pair → rebuild fails after clearing the anchor → rollback restores the signed document without restoring its anchor → edit the existing UTXO amount → the old 02 signature survives and export is enabled. The same bypass applies to the other signing types.

Pair decorations are excluded, but doc.tx decorations are not: changing only output asm changes the anchor with identical serialized bytes.

MuSig2 type assignments match BIP-373. BIP-327 permits nonce generation before the message is known, so dropping 0x1b is conservative session cleanup.

src/js/psbt-editor.js:786-795 anchor cleared before the fallible build; re-anchored only on success.
src/js/psbt-editor.js:886-896 rollback restores signatures without their anchor.
src/js/psbt-editor.js:277-310 drop set and UTXO coverage; whole doc.tx includes decorative fields.
test/psbt-editor-tx.test.mjs:140-184 helper/wiring coverage exists, misses rollback recovery.

Does not close #325.

portlandhodl:
Both findings reproduced exactly as described (state-machine simulation driving the real `signingAnchor`/`dropSigningPairs`/`psbtBuildBytes`): the duplicate-0x01 rollback left the restored signature anchorless, and the follow-up UTXO-amount edit kept 0x02 with export enabled; an asm-only edit moved the anchor with identical serialized bytes.

Fixed in 4a3cf3a:

- **Rollback** — `mutate()` snapshots `pristineTx` with the document and restores both together, so the anchor can no longer be orphaned by a rejected structural edit.
- **Decorative tx fields** — `signingAnchor` now keys the transaction through `psbtEditorBuildDoc(doc).tx`, the same projection the builder reads; presentation fields (output asm, like pair name/decoded) cannot move it.

Coverage added in `test/psbt-editor-tx.test.mjs`: rollback recovery (duplicate-0x01 rejection keeps anchor + signature; the next UTXO edit still drops it), decorative-field stability, and wiring assertions pinning the snapshot/restore.

With the bypass closed this should now close #325. `npm run build && npm test` green (1156 pass, 1 engine-skip).

portlandhodl:
Update after merging current `rock` (898c810): the rollback fix landed there independently via #383 — as a superset that also restores the `stale` flag — so `mutate()` resolves to rock's version and this PR now carries only:

- `signingAnchor(doc)` = build-projection tx + each input's UTXO declarations (0x00/0x01) — UTXO-claim edits trip the same `dropSigningPairs` path as transaction edits, and decorative fields (output asm) cannot move the anchor;
- `SIGNING_KEY_TYPES` += BIP-327 MuSig2 `0x1b`/`0x1c` (participant pubkeys `0x1a` stay);
- tests: anchor coverage/stability, decorative-field stability, MuSig2 drop semantics, and a behavioral rollback test asserting a post-rollback UTXO-claim edit still drops the signature (the exact bypass above, now closed end-to-end on the merged result).

`npm run build && npm test` green on the merge (1183 pass, 1 engine-skip).

MrHodlX (approved):
ACK

The prior NACK no longer stands on 898c810. Actual mutate()/rebuild() execution confirms duplicate 01 rejection restores the document, signing anchor, and stale flag; the subsequent UTXO-amount edit removes 02 from export bytes. The anchor uses projected transaction fields plus only 00/01 key/value pairs; output-asm-only changes leave both anchor and bytes unchanged. Witness amount/scriptPubKey, non-witness prevtx, and transaction edits drop 02/07/08/13/14/1b/1c; 03 edits preserve signatures and 1a stays. The delta is editor plus tests, with no dependency or HTML changes. All 15 editor tests pass; on unpatched rock, the new tests fail at the missing export, and an independent behavioral check reproduces signature retention. Build passes; full suite has 1,176 passes, four skips, and one failure from no installed browser. This fixes the reported bypass, but does not establish complete #325 acceptance: editor regression coverage still lacks the explicitly requested ECDSA/Taproot sighash variants.

src/js/psbt-editor.js:928–948: rollback restores anchor and stale state.
src/js/psbt-editor.js:255–313, 819–828: build projection, drop set, UTXO anchor, rebuild enforcement.
test/psbt-editor-tx.test.mjs:211–253: duplicate-01 rollback regression; follow-up edits scriptPubKey bytes rather than amount.
test/psbt-editor-tx.test.mjs:131–208, 256–264: rollback wiring, signing types, decorative-field stability, and anchor wiring.

DOES NOT CLOSE: #325

Issue #283, opened 2026-09-02
🐛🐛🐛 bug: Wrong keys derived when adding numbers as passphrase


dani69654:
### Summary:

ColdCard allows you to set a bunch of numbers as passphrase.
Once those numbers are added as passphrase to both ColdCard AND EntropyLab, those two show a different fingerprint, therefore deriving different keys.

### Expected behavior:

The generated fingerprint should be the same on both devices.

hotpixelgroup:
Thanks for the report — a silent derivation mismatch with a hardware signer is about the worst failure mode this tool could have, so I dug into this thoroughly. I have not been able to reproduce it on current `rock`, and I would like your help narrowing down where your divergence came from.

**What I verified** (real assembled page in headless Chromium with trusted input events, compared against an independent implementation, `@scure/bip39` + `@scure/bip32`). With the standard test mnemonic, numeric passphrases produce exactly the expected master fingerprints:

| Passphrase | Master fingerprint |
|---|---|
| *(empty)* | `73c5da0a` |
| `123456` | `e9646cf9` |
| `007` | `9f67e695` |
| `0` | `4d9cb38e` |
| `987654321098765432109876543210` | `aa5d8432` |

Every passphrase-bearing flow I could drive matched the independent implementation byte-for-byte: typed input → live fingerprint preview, the full *Derive Key* flow, the on-screen keyboard's numeric mode (character codes verified), Key Station tab switching/restore, an encrypted Key Manager (`.elkeys`) export/re-import round-trip (leading zeros survive), BIP85 parent root and child words, multisig cosigner xpubs, and the vanity grind's *Update key* write-back. A static audit of every value path between the input field and PBKDF2 found no place where a numeric string could be coerced, trimmed, or altered — leading zeros like `007` are preserved everywhere.

I have opened PR #376 adding a regression guard that pins these numeric-passphrase fingerprints and asserts the pairs a coercion bug would collapse (`007`/`7`, `0`/`00`, trailing whitespace) stay distinct, so this class of bug cannot slip in later.

**To find what happened in your case, could you share:**

1. The exact steps you took in EntropyLab — which entry method you used for the seed words (seed phrase field, dice, import, …) and where you typed the passphrase?
2. The version/commit shown in the page footer when you saw the mismatch?
3. The passphrase itself, if you are comfortable sharing it (or at least its shape: length, leading zeros, any spaces or symbols)?
[...]

PR #376, opened 2026-09-05
test: guard numeric BIP39 passphrases against exact-string regressions (#283)

hotpixelgroup:
Refs #283.

## Summary

Issue #283 reports that a BIP39 passphrase made of digits produces a different master fingerprint in EntropyLab than on a ColdCard. I investigated exhaustively and **could not reproduce any divergence**: every passphrase-bearing flow in the page stretches the exact string typed and derives keys identical to an independent implementation. This PR adds a regression guard that locks that behavior in, so a future change that coerces, trims, or caps a numeric passphrase fails loudly instead of silently deriving an unrecoverable wallet.

## Verification performed (no divergence found)

All checks compared the real assembled page (headless Chromium, trusted input events) against an independent oracle (`@scure/bip39` + `@scure/bip32`, and `node:crypto` for the BIP85 HMAC). Anchor vectors with mnemonic `abandon … about` (12 words):

| Passphrase | Expected master fingerprint | Observed |
|---|---|---|
| *(empty)* | `73c5da0a` | MATCH |
| `123456` | `e9646cf9` | MATCH |
| `007` | `9f67e695` | MATCH |
| `0` | `4d9cb38e` | MATCH |
| `987654321098765432109876543210` | `aa5d8432` | MATCH |
| `abc123` | `23a41e58` | MATCH |
| `hello world` | `ec372958` | MATCH |

Flows exercised, all byte-exact against the oracle:

- Seed field → typed passphrase (real `Input.insertText` events) → debounced fingerprint preview
- Full *Derive Key* flow — wallet result fingerprint
- On-screen keyboard numeric mode, digit by digit (field char codes verified)
- Key Station tab switch / restore — `#pass` value byte-exact (leading zeros intact)
- Encrypted Key Manager (`.elkeys`) export → in-memory copies removed → re-import → passphrase restored byte-exact, re-derive matches
- BIP85: passphrased parent root xprv byte-identical to oracle; 12-word English child at index 0 identical to an independent BIP85 implementation, for both passphrases
- Multisig cosigner strings (xpub + fingerprint) byte-identical to oracle for a 1-of-2 Native SegWit setup
- Vanity passphrase grind → *Update key* write-back — passphrase string preserved byte-exact, re-derives to the oracle fingerprint

A static audit of every value path (input listener, preview, calculate, global sync, autocomplete, input filters, on-screen keyboard layouts, serialization) found no numeric coercion anywhere.

## What the test locks

New test `BIP39 numeric passphrases stretch the exact string (#283)` in `test/hdkey-bip39-wasm.test.mjs`:

- 7 passphrase cases pinned to hardcoded master fingerprints produced by the independent implementation, so the guard survives oracle dependency changes
- A seed-level differential against `@scure/bip39` for each case
- Distinctness assertions for the pairs a numeric coercion would collapse: `007`/`7`, `0`/`00`, `000102`/`102`, and `123456` vs a trailing-space variant

## Testing

- `npm run build` — clean
- `npm test` — 1118 pass, 0 fail (6 skips: browser engines not installed locally)
- `npm run verify` — site artifact verified

## Note on the issue

Since no divergence reproduces on current `rock`, the remaining possibilities are an older release, a different flow than the ones above, or the entry side on the ColdCard. I have asked the reporter on #283 for their exact reproduction path; happy to follow up there if new details surface.

### DIFF
--- a/test/hdkey-bip39-wasm.test.mjs
+++ b/test/hdkey-bip39-wasm.test.mjs
@@ -245,6 +245,38 @@ test("BIP39: no non-canonical separator yields a derivable seed-bearing phrase",
   }
 });
 
+// Regression guard for #283 ("Wrong keys derived when adding numbers as
+// passphrase"): a passphrase of ASCII digits must reach PBKDF2 as the exact
+// string typed — no numeric coercion (leading zeros are significant), no
+// trimming, no length capping. ColdCard and every other BIP39 signer stretch
+// the string as-is, so any transformation here derives a wallet nothing else
+// reproduces. The fingerprints are pinned to values an independent
+// implementation (@scure/bip39 + @scure/bip32) produces, so the guard holds
+// even if the oracle dependency moves.
+test("BIP39 numeric passphrases stretch the exact string (#283)", () => {
+  const phrase = "abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon about";
+  const fingerprintHex = (pass) => (HDKey.fromMasterSeed(mnemonicToSeedSync(phrase, pass)).fingerprint >>> 0).toString(16).padStart(8, "0");
+  const cases = [
+    ["", "73c5da0a"],
+    ["0", "4d9cb38e"],
+    ["007", "9f67e695"],
+    ["123456", "e9646cf9"],
+    ["987654321098765432109876543210", "aa5d8432"],
+    ["abc123", "23a41e58"],
+    ["hello world", "ec372958"],
+  ];
+  for (const [pass, expected] of cases) {
+    assert.equal(bytesToHex(mnemonicToSeedSync(phrase, pass)), bytesToHex(scureBip39.mnemonicToSeedSync(phrase, pass)), `seed for ${JSON.stringify(pass)} disagrees with scure`);
+    assert.equal(fingerprintHex(pass), expected, `master fingerprint for ${JSON.stringify(pass)}`);
+  }
+  // Numeric coercion would collapse these pairs; exact-string semantics keep
+  // every one distinct.
+  const distinct = [["007", "7"], ["123456", "123456 "], ["0", "00"], ["000102", "102"]];
+  for (const [a, b] of distinct) {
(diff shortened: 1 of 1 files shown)


Hero-Gamer (commented):
Useful exact-string lock on the seed path; it does not cover the UI path #283 described, and #283 is already closed as unreproducible.

The new test in `test/hdkey-bip39-wasm.test.mjs` calls `mnemonicToSeedSync` / `HDKey.fromMasterSeed` with string literals. That will fail if the WASM/JS wrapper coerces, trims, or caps a numeric passphrase, and the distinctness pairs (`007`/`7`, `0`/`00`, `000102`/`102`, trailing space) are the part that would actually catch a numeric collapse. Seed-vs-scure plus pinned fingerprints is the right shape for this layer.

It will not fail if something between `#pass` / the on-screen keyboard / restore and PBKDF2 rewrites the string. The PR body reports those flows were driven by hand; they are not in the diff. Existing TREZOR and scure differentials already cover “some passphrase reaches PBKDF2.”

#283 was closed by the reporter after they could not reproduce either. This PR should not be treated as closing that issue. Rebase onto current `rock` before merge — base here is still `226ed9b` and mergeable is unstable.

Not requesting a UI harness on this PR. If a later change wants to lock the reported failure mode, that belongs in an input/restore test, not in this wrapper file.

w-s-bitcoin:
Thanks for the thorough investigation and for documenting the paths you checked. Issue #283 could not be reproduced on current EntropyLab, and the reporter subsequently confirmed they could not reproduce it either and closed the issue.\n\nEntropyLab should handle passphrases according to BIP-39 itself: apply the required Unicode normalization and use the resulting passphrase without device-specific interpretation, numeric coercion, trimming, or length rules. The current suite already covers the published BIP-39 passphrase vector, normalization, and differential seed derivation.\n\nAlthough this PR only adds tests and does not change passphrase behavior, its additional coverage is tied to an unconfirmed hardware-wallet mismatch and does not exercise the UI path described by the original report. We are therefore closing it rather than carrying a device-specific regression guard for a bug that was not demonstrated. Thanks again for taking the security report seriously.

PR #378, opened 2026-09-05
Hand SP sender keys around as zeroable bytes, and wipe them after use

portlandhodl:
## Problem

Follow-up to #331. `hodlSpDeriveVinKeys` verified and wiped the derived hdkey node — and simultaneously returned the same scalar as `private_key: hodlSpBytesToHex(node.privateKey)`, an immutable JS string that no `wipePrivateData()` can touch. The derive-verify-wipe story was therefore cosmetic for the scalar itself, and `eligibleInputKeys` (src/js/bip352.js) then materialised it again as `hexToBytes` output plus BigInt scalars.

## Change (src/js/app.js only)

- The derived key leaves the resolver as `new Uint8Array(node.privateKey)` — a zeroable copy made before the node wipe — instead of a hex string. (`eligibleInputKeys` accepts bytes, which also removes the `hexToBytes` re-materialisation.)
- `hodlRenderSpSend` runs `createSilentPaymentOutputs` inside `try/finally` so the new `hodlSpWipeVinKeys` zeroes every derived copy on the success and error paths alike.
- The resolver comment now states the residual honestly: the BigInt scalars inside the BIP-352 math are immutable JavaScript values and cannot be wiped; they live until GC. (The library keeps returning `inputPrivateKeySum` for the published-vector API; the UI flow never reads it.)

## Tests (test/sp-send-session.test.mjs)

- The resolver now must return 32-byte `Uint8Array` keys (deep-equal across the index/path/fingerprint resolution paths — the previous string-equality assertions are updated).
- New: `hodlSpWipeVinKeys` zeroes derived byte copies and skips non-byte entries; the `try { createSilentPaymentOutputs … } finally { hodlSpWipeVinKeys(keyedVins) }` wiring is pinned at source level.

## Ran

`npm run build`, `npm test` — green apart from two pre-existing headless-Firefox journal layout flakes ("Journal notepad…", "Journal embeds a Key Station LifeHash…") that fail intermittently in this environment on clean `rock` too and pass in isolated runs; neither touches the SP code paths changed here.

### DIFF
--- a/src/js/app.js
+++ b/src/js/app.js
@@ -9128,11 +9128,14 @@ function hodlRenderSpReceive() {
 }
 // Resolve every eligible vin's input key from the loaded SP session root —
 // never from a pasted scalar (issue #331). A vin may name its derivation
-// path explicitly ("path": "m/84'/1'/0'/0/5", plus an optional "fingerprint"
+// path explicitly ("path": "m/84'/1'/0'/0/0/5", plus an optional "fingerprint"
 // that must equal the session root's) or be resolved through the session
 // ownership index by its prevout script. Either way the derived key is
-// verified against the prevout script before use, and every derived node is
-// wiped after output construction.
+// verified against the prevout script before use, every derived node is
+// wiped immediately, and the scalar leaves here only as a zeroable byte copy
+// that the send flow wipes after output construction (hodlSpWipeVinKeys).
+// What no wipe can cover: the BigInt scalars the BIP-352 math materialises
+// are immutable JavaScript values and live until the GC takes them.
 function hodlSpDeriveVinKeys(vins) {
   hodlSpEnsureHd();
   const root = hodlSpHd, network = hodlSpNetwork(), fingerprint = hodlFingerprintHex(root.fingerprint);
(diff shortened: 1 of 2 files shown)


MrHodlX (changes requested):
BLOCK.

The resolver now returns 32-byte Uint8Array copies, and output construction wipes them on success or throw. But resolution itself runs outside try/finally: a valid first input followed by a foreign-fingerprint input leaves the first copy nonzero. The "no scalar hex strings" claim also fails: BIP-352 converts bytes through hex and returns a hex inputPrivateKeySum even when unread by the UI. The BigInt comment is accurate but incomplete. This improves part of #331's cleanup requirement; it does not complete it.

P2TR still returns the untweaked internal scalar. Reproduced an output the recipient scanner did not find, contrary to BIP-352 sender/receiver agreement.

Dirty mergeability independently blocks shipment; conflicts exist in both changed files.

src/js/app.js:9186-9196 byte-copy handoff, node wipe, and copy-zeroing helper.
src/js/app.js:9143-9155, 9206-9211 partial-resolution failure bypasses cleanup.
src/js/bip352.js:27-32, 383, 473 resolver bytes bypass the string-input branch, but downstream scalar hex remains.
src/js/app.js:9179-9186 verifies tweaked P2TR script, returns untweaked key.
test/sp-send-session.test.mjs:83-111 requested assertions exist; partial-resolution cleanup is untested.

Does not close #331.

Issue #340, opened 2026-09-05
Require exact CompactSize count values in PSBT v2 globals


Rob1Ham:
## Priority
Dormant parser correctness blocker

## Summary
The unused PSBT v2 helper decodes global input/output counts with `hodlVarInt(...)[0]` but does not verify that the CompactSize value consumes the entire PSBT value. A valid count followed by trailing bytes is accepted by this helper even though the typed field is malformed.

## Affected code
- `src/js/psbt-v2.js:18-22` discards the returned end offset
- `src/js/psbt-v2.js:30-36` trusts the resulting count

## Acceptance criteria
- require the CompactSize decoder end offset to equal the field value length
- reject noncanonical encodings and all trailing bytes
- add tests for one-byte counts plus trailing data, multi-byte canonical counts, truncated values, and noncanonical encodings
- keep this as a blocker before wiring PSBT v2 into runtime

Issue #320, opened 2026-09-05
Disable PSBT exports while displayed edits are stale


Rob1Ham:
## Severity
Medium wrong-transaction workflow risk

## Summary
When a live PSBT field edit fails to build, the editor leaves the new invalid value visible but retains the previous valid PSBT bytes. Copy, download, reload, and QR controls remain active and continue exporting those old bytes.

The warning and dimmed styling help, but the mismatched artifact remains actionable, including across an air-gapped QR boundary.

## Affected code
- `src/js/psbt-editor.js:690-718` retains `resultBytes` after failure
- `src/js/psbt-editor.js:568-610` binds artifact actions regardless of stale state
- `src/js/psbt-editor.js:614-635` continues encoding old bytes as QR
- `test/psbt-live-qr.test.mjs:55-67` tests stale styling but not action blocking

## Impact
A user can believe an address, amount, outpoint, sequence, or locktime was corrected while sending the preceding transaction to a signer.

## Acceptance criteria
- copy, download, reload, and QR output are unavailable while stale
- no stale bytes can cross an export boundary
- browser regression test covers a valid edit followed by an invalid pending value

PR #379, opened 2026-09-05
Name the gate that actually ran in the PSBT editor's rebuild verdict

portlandhodl:
## Problem

The result panel printed "Rebuilt PSBT parses under rust-bitcoin; its unsigned transaction passes consensus sanity checks" for **every** build. For PSBT v2 that first clause is false: rust-bitcoin's `Psbt` type is v0-only (deprecated upstream), so the v2 build path (`build_v2` in psbt-wasm/src/lib.rs) is gated by the crate's own BIP-370 reader — a closed-loop parse + transaction round-trip — and the inspect side already reports `rustBitcoinError: null` for v2 precisely because the rust-bitcoin check never ran. The header verdict made this distinction (issue #358); the result line contradicted it.

## Change

src/js/psbt-editor.js `renderResult`: the verdict sentence now branches on `doc.psbtVersion` — v2 builds say "Rebuilt PSBT v2 round-trips through EntropyLab's own BIP-370 reader (rust-bitcoin checks v0 only)", v0 builds keep "Rebuilt PSBT parses under rust-bitcoin". The consensus-sanity clause applies to both versions (the build gate runs `tx_sanity_error` before either serializer) and is unchanged.

## Test (test/psbt-live-qr.test.mjs)

A v2 document built through the real WASM facade has `psbtVersion === 2` and `rustBitcoinError === null` (the null means "never ran", not "accepted"), and the source-level assertion pins the v2/v0 verdict branch. Verified the test fails without the fix.

## Ran

`npm run build`, `npm test` — green apart from two pre-existing headless-Firefox journal layout flakes that fail intermittently in this environment (clean `rock` included) and pass in isolated runs; neither touches the editor.

## Note

This and the #320 stale-blanking PR both touch `renderResult`; whichever merges second will see a trivial two-line conflict.


MrHodlX (approved):
SHIP.

This fixes the v2 copy lie only. v2 names EntropyLab's BIP-370 reader; v0 credits rust-bitcoin. Both match their executed build gates. On v2, rustBitcoinError: null means the rust-bitcoin check never ran. The closed-loop reader is mandatory on this path; if skipped, the new sentence would be false.

Both versions run the shared sanity gate, but this SHA does not include 373's null-prevout check — null prevouts remain accepted here. That qualification prevents claiming complete consensus validation. V0 carrying v2-only fields remains accepted.

src/js/psbt-editor.js:633-638 v2/v0 verdict branch and unchanged sanity clause.
psbt-wasm/src/lib.rs:1132-1141 v2 null means "not run."
psbt-wasm/src/lib.rs:1304-1322, 1370, 1477-1481 shared sanity gate, v0 deserialize, mandatory v2 round-trip.
test/psbt-live-qr.test.mjs:84-96 version, null-result, and wording assertions.

Does not close #340, #358.

MrHodlX (approved):
ACK

The prior SHIP stands on 5c22ae6. The verdict fix and regression test are unchanged after rebasing. V2 credits EntropyLab’s BIP-370 reader and explicitly limits rust-bitcoin checks to v0; v0 still credits rust-bitcoin. V2’s rustBitcoinError === null means never ran, not accepted. Both versions execute tx_sanity_error, but null prevouts remain accepted, so this is not complete consensus validation. 5c22ae6 (“Rebuild site artifact [skip ci]”) changes only entropylab.html, which matches a source-parent rebuild byte-for-byte: no additional editor logic or dependencies, and no HTML edit in the source commit. All nine live-QR tests pass; the verdict test fails on unpatched rock. Full suite: 1,172 passed, four skipped, one failure because no supported browser is installed. This fixes only the result-line lie.

src/js/psbt-editor.js:635–640: version branch and wording.
psbt-wasm/src/lib.rs:1132–1142: v2 null means check skipped.
psbt-wasm/src/lib.rs:1035–1061, 1304–1322, 1370, 1477–1481: sanity limitations, shared gate, v0 deserialize, mandatory v2 round-trip.
test/psbt-live-qr.test.mjs:98–110: version, null-result, and wording assertions.

DOES NOT CLOSE: #340, #358

PR #380, opened 2026-09-05
Blank the stale PSBT bytes, not just the export controls (issue #320)

portlandhodl:
## Problem

Follow-up to #320. While the editor fields do not build, the result panel shows the last valid build with every export control disabled — but the base64/hex textareas were left populated and merely `readonly` + `disabled`. That is not an export boundary: **Firefox lets a disabled, readonly textarea's content be selected and copied** (long-standing Gecko behaviour), so the stale bytes stayed one drag-and-copy away from leaving the page while the UI said export was unavailable.

## Change (src/js/psbt-editor.js)

- `renderResult`: while stale, both textareas render empty (`${stale ? "" : escapeHtml(b64)}` / hex) — the bytes never enter the DOM.
- `markResultStale` (the keystroke path, where the panel was rendered non-stale): blanks both textareas' live values alongside the existing disabling.
- A clean rebuild refills both as before; comments updated to say why disabling alone is insufficient.

## Tests

- test/psbt-live-qr.test.mjs: the existing #320 source-level case now also pins the two template blankings and the `area.value = ""` keystroke-path blanking. Verified it fails without the fix.
- test/browser-suite.html: the end-to-end stale-flow case (type an invalid value mid-edit) now asserts both textareas read empty while stale, ahead of the existing rebuild-and-recover assertions.

## Ran

`npm run build`, `npm test`, `npm run test:browser` — green apart from two pre-existing headless-Firefox journal layout flakes ("Journal notepad…", "Journal embeds a Key Station LifeHash…") that fail intermittently in this environment on clean `rock` too and pass in isolated runs; neither touches the editor.

## Note

This and the v2-verdict PR both touch `renderResult`; whichever merges second will see a trivial two-line conflict.

### DIFF
--- a/src/js/psbt-editor.js
+++ b/src/js/psbt-editor.js
@@ -622,21 +622,23 @@ export const initPsbtEditor = ({ networkDefault = () => "mainnet" } = {}) => {
     const b64 = base64Encode(resultBytes);
     const hex = bytesToHex(resultBytes);
     box.classList.toggle("psbted-stale", stale);
-    // While the displayed fields do not build, the last valid bytes stay
-    // visible for reference but must not cross an export boundary: every
-    // copy/download/reload control and the QR are disabled (issue #320).
+    // While the displayed fields do not build, the last valid build must not
+    // cross an export boundary: every copy/download/reload control and the QR
+    // are disabled — and the byte text is blanked, because a disabled,
+    // readonly textarea's content is still selectable and copyable in Firefox
+    // (issue #320).
     const gated = stale ? " disabled" : "";
     box.innerHTML = `
       ${stale ? `<p class="psbted-note-warn" id="psbted-stale-note">The fields do not build right now — this is the last valid build. Export is unavailable until they build again.</p>` : ""}
       <p class="psbt-ok">Rebuilt PSBT parses under rust-bitcoin; its unsigned transaction passes consensus sanity checks (${resultBytes.length} bytes).</p>
-      <label class="field">Edited PSBT (base64)<textarea id="psbted-result-b64" readonly spellcheck="false"${gated}>${escapeHtml(b64)}</textarea></label>
+      <label class="field">Edited PSBT (base64)<textarea id="psbted-result-b64" readonly spellcheck="false"${gated}>${stale ? "" : escapeHtml(b64)}</textarea></label>
       <div class="row psbt-actions">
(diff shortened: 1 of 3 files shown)


MrHodlX (changes requested):
BLOCK.

This closes the visible Firefox copy hole, but fails the requested DOM/export boundary. markResultStale() clears .value while leaving last-good bytes in both textareas' .textContent and .defaultValue. Existing handlers also remain callable: a synthetic click exported last-good base64 in the DOM reproduction; normal disabled-button clicks were blocked. QR clears, and a successful rebuild restores both new encodings and re-enables export. No new dependency. No hand-edited entropylab.html.

Build and the focused tests pass; the patched source test fails against unpatched rock. Browser assertions exist but do not cover retained DOM text or callable handlers.

src/js/psbt-editor.js:634-641 stale full-render blanking.
src/js/psbt-editor.js:703-718 live values cleared, controls disabled, QR removed; textarea text nodes remain.
src/js/psbt-editor.js:647-661 retained export handlers lack stale guards.
src/js/psbt-editor.js:775-783 fresh bytes and export recovery.
test/psbt-live-qr.test.mjs:84-86 and test/browser-suite.html:2473-2480 cover blank values and recovery, not retained DOM text or callable handlers.

Does not close #320.

MrHodlX (approved):
SHIP

At 689e807, the prior BLOCK is resolved: stale textareas have empty .value, .textContent, and .defaultValue; all four export handlers reject synthetic clicks; QR clears; successful rebuild restores both encodings, QR, and enabled exports. The DOM reproduction found no retained last-good payload or remaining stale export path. This addresses #320 beyond the Firefox copy hole. Focused tests pass 8/8 and fail on unpatched rock. Build passes; full tests: 1,145 passed, 4 skipped, 1 failure because no browser is installed. Changes cover only the editor and its tests—no dependency or generated HTML changes.

src/js/psbt-editor.js:634,641 — stale template blanking.
src/js/psbt-editor.js:650–666 — four handler guards.
src/js/psbt-editor.js:714–737 — disable controls, clear values and DOM text, remove QR.
src/js/psbt-editor.js:793–800 — successful rebuild recovery.
test/psbt-live-qr.test.mjs:84–94; test/browser-suite.html:2473–2489 — retained-text and handler regressions.

DOES NOT CLOSE: #325

Issue #333, opened 2026-09-05
Correct and wire Taproot signature sighash parsing


Rob1Ham:
## Priority
Medium before full Taproot signing support

## Summary
The existing Schnorr helper has two encoding errors and is not wired into the active PSBT inspector:

- a valid 65-byte Schnorr signature is rejected whenever its first signature byte is `0x30`
- a 65-byte signature ending in explicit `0x00` is treated as `SIGHASH_DEFAULT`, although DEFAULT must be represented by the 64-byte form and an explicit zero suffix is invalid
- active inspection counts Taproot signature fields but does not decode their sighash bytes into the safety analysis

## Affected code
- `src/js/psbt-schnorr.js:3-13` applies the incorrect first-byte and suffix rules
- `src/js/psbt-schnorr.js:17-25` describes all non-ALL policies as not committing to outputs, including ANYONECANPAY combinations that still commit to all outputs
- `src/js/app.js:7991-7993` and `9386-9400` count Taproot fields without parsing suffixes
- `test/psbt-schnorr-nonce.test.mjs:40-48` currently expects rejection based on first byte `0x30`

## Acceptance criteria
- accept exactly 64-byte signatures or valid 65-byte signatures with a nonzero defined Taproot sighash byte
- reject explicit suffix `0x00`
- never classify Schnorr versus DER based on the first signature byte
- wire Taproot key/script signature parsing into active inspection
- distinguish missing input commitments from missing output commitments in warnings
- add BIP340/BIP341 vectors covering DEFAULT, ALL, NONE, SINGLE, ANYONECANPAY, first byte `0x30`, and explicit zero

PR #381, opened 2026-09-05
Count only parseable Schnorr signatures as present (issue #333)

portlandhodl:
## Problem

Follow-up to #333. The parser deliberately keeps unparseable Schnorr values in `tapSignatures` (with `r: null`) so the policy loop can flag them — "an unparseable Schnorr signature is flagged, not silently counted" (app.js). But the count lines ran on the raw collection length:

- the per-input header: `signatures.length + tapSignatures.length + " signature(s) present"`;
- the totals: `tapSignatureCount += tapSignatures.length`, feeding "This PSBT also contains N Taproot / Schnorr signature(s)" and the nonce-analysis completeness gate.

A 32-byte garbage value in a `PSBT_IN_TAP_KEY_SIG` field was thus counted as a present signature *and* flagged as "not valid under BIP341" — the count line contradicted the warning.

## Change (src/js/app.js)

Both count sites now run on the parseable subset: `parsedTapSignatures = tapSignatures.reduce((count, tapSig) => count + (tapSig.r ? 1 : 0), 0)`. An input whose only Taproot sig is unparseable now reads "Not signed yet" plus the policy problem, instead of "1 signature(s) present" plus the policy problem. The unparseable entry still feeds the incomplete-policy verdict through the existing `uninspected` path; `policyIncomplete` handling is untouched.

## Tests (test/psbt-schnorr-nonce.test.mjs)

- Unit: a 64-byte and a 10-byte `PSBT_IN_TAP_KEY_SIG` together — the collection keeps both records, exactly one has `r`.
- Source-level: the two count sites are pinned to `parsedTapSignatures`, and the old `tapSignatureCount += tapSignatures.length` form is asserted gone. Verified the test fails without the fix.

## Ran

`npm run build`, `npm test` — green apart from two pre-existing headless-Firefox journal layout flakes ("Journal notepad…", "Journal embeds a Key Station LifeHash…") that fail intermittently in this environment (clean `rock` included) and pass in isolated runs; neither touches the PSBT report.

### DIFF
--- a/src/js/app.js
+++ b/src/js/app.js
@@ -9562,8 +9562,12 @@ function hodlRenderPsbt(psbt) {
       signatures = signatures.concat(finalMaterial.signatures);
       uninspected += finalMaterial.uninspected + (finalMaterial.malformed ? 1 : 0);
     }
-    tapSignatureCount += tapSignatures.length;
-    html.push("<p class='psbt-kv'><strong>Input " + index + "</strong> \xB7 " + hodlHexRev(previous.txid) + " : " + previous.vout + (claim ? " \xB7 " + hodlSats(claim.amount) + " BTC claimed" : "") + "<br>" + hodlEscapeHtml(destination) + "<br>" + (signatures.length + tapSignatures.length ? signatures.length + tapSignatures.length + " signature(s) present" : finalized ? "Finalized input data present" : "Not signed yet") + (declaredSighashError ? "<br>Declared sighash policy unreadable: " + hodlEscapeHtml(declaredSighashError) : "<br>Signature policy: " + hodlEscapeHtml(declaredLabel)) + "</p>");
+    // Only a Schnorr signature that parses under BIP341 counts as present: an
+    // unparseable one is flagged as a policy problem below, not silently
+    // counted (issue #333).
+    let parsedTapSignatures = tapSignatures.reduce((count, tapSig) => count + (tapSig.r ? 1 : 0), 0);
+    tapSignatureCount += parsedTapSignatures;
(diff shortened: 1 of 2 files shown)


MrHodlX (approved):
SHIP.

PR 381 (one commit f5f67f8, two files) does exactly the count-line fix and nothing else. Both count sites now use the parseable subset: parsedTapSignatures is computed at src/js/app.js:9568 (tapSignatures.reduce, tapSig.r), the Taproot total at :9569 (tapSignatureCount += parsedTapSignatures), and the per-input line at :9570 (signatures.length + parsedTapSignatures). An input whose only Taproot field is a 32-byte garbage TAP_KEY_SIG falls through to "Not signed yet". I confirmed no other site still reads tapSignatures.length as a count.

The unparseable record is still collected with r: null by hodlTapKeySigs (src/js/psbt-schnorr.js:49-56) and still hits the policy loop (app.js:9592-9598 uninspected / policyIncomplete / policyProblems, including the "not valid under BIP341" warning). ECDSA and finalized paths are byte-identical and not in this diff.

test/psbt-schnorr-nonce.test.mjs:142-159 builds a 64-byte and a 10-byte TAP_KEY_SIG, asserts both records survive and exactly one has r, then pins the three count expressions. The old tapSignatureCount += tapSignatures.length form is gone.

This closes only the count lie. #333's parser work already landed in 372. Fabu's #328 blocker still stands: a one-byte 0x01 PSBT_IN_PARTIAL_SIG is still counted as "1 signature(s) present" by the inspector (app.js:8048-8056 / :9570) and as "1 signature" on the diagram, because the WASM decoder only rejects an empty partial sig. Explicit 0x00 / undefined Taproot suffix still shows as a signature on the diagram while the inspector rejects it.

Does not close #328, #331, #340, #358, #364.

PR #382, opened 2026-09-05
feat(#191): expose PSBT UTXO verification states

### LINKED ISSUE #191: Add independent PSBT UTXO verification

Hero-Gamer:
## Summary

This PR implements the read-only UI portion of #191 by exposing the verification state of PSBT input UTXO amounts in the existing PSBT visualizer.

The visualizer now distinguishes between:

- **Verified** — a matching non-witness UTXO is present and has been validated by the existing PSBT inspection pipeline.
- **Unverified** — only a witness UTXO claim is available, or there is no usable UTXO declaration.
- **Mismatch** — valid witness and non-witness UTXO amount claims disagree.

This builds on the UTXO validation and conflict detection already implemented in the PSBT inspector, rather than introducing a second validation path in JavaScript.

## What changed

- Added UTXO verification-state classification to the PSBT visualizer.
- Display verification status alongside input amounts.
- Preserve the existing fee and input-conflict behavior.
- Added focused tests covering the verification states and relevant edge cases.
- Kept the implementation read-only and offline.

## Verification semantics

A non-witness UTXO is treated as verified only when the existing inspector exposes its decoded `prevout`. The Rust PSBT inspection code already validates the supplied transaction and its relationship to the input outpoint before exposing that information.

The UI therefore does not perform its own transaction parsing, txid validation, vout validation, or network lookup.

A witness UTXO by itself remains **unverified** because it provides a claim about the previous output rather than the independently supplied previous transaction used for non-witness verification.

If both witness and non-witness declarations are valid but their amounts disagree, the input is shown as **mismatch**, preserving the existing behavior that prevents a definitive fee calculation.

## Tests

Added coverage for:

- matching non-witness UTXO → **Verified**
- witness-only UTXO → **Unverified**
- no usable UTXO declaration → **Unverified**
- malformed non-witness UTXO → **Unverified**
- non-witness declaration without a validated `prevout` → **Unverified**
- agreeing witness and non-witness claims → **Verified**
- conflicting valid witness and non-witness claims → **Mismatch**

## Scope

This PR is intentionally limited to exposing verification state in the existing PSBT visualization flow.

Out of scope:

- network or chain lookups
- signing or custody functionality
- changes to PSBT parsing or fee calculation
- transaction broadcasting
- a second JavaScript UTXO-validation implementation
- scriptPubKey comparison between witness and non-witness UTXO declarations

Closes #191


Hero-Gamer:
### Implementation clarification — current diff (supersedes PR body Summary/Verification)

The PR body predates the final implementation. The current `psbt-viz.js` does:

**Verification requires both witness and validated non-witness claims to agree on amount and scriptPubKey.**

* **Verified** — validated non-witness prevout + witness UTXO present, with matching amount **and** scriptPubKey. The Rust inspector already validates the non-witness transaction's txid/vout before exposing the prevout.
* **Unverified** — only one side is present, malformed, unmatched, or there is no usable declaration.
* **Mismatch** — both claims are present but their amount **or** scriptPubKey disagrees (including equal amounts with different scripts).

**Fee:** When all inputs are Verified, the fee is derived from the sum of independently verified input amounts minus transaction outputs and is labeled `(independently verified)`. Otherwise, the existing PSBT-claim fee behavior is preserved. This is intentional per #191.

**Boundary — UI projection only, no second validator:**
The JS layer consumes the existing WASM/inspector result. It does not independently decode or validate the previous transaction a second time.

Conceptually, the relevant data contract is:

```json
{
  "witness_utxo": {
    "amount": "u64",
    "scriptPubKey": "hex"
  } | null,
  "non_witness_prevout": {
    "amount": "u64",
    "scriptPubKey": "hex"
  } | null,
  "validated": true
}
```

src/js/psbt-viz.js
```diff
@@ -101,9 +150,15 @@ const signingStatus = (pairs) => {
   return { text: "unsigned", tone: "muted" };
 };
 
-// sats null with every input claimed marks an invalid fee: the document's
-// error string says why (negative fee, u64 overflow, or past MAX_MONEY).
+// Render the fee summary while preserving the inspector's unknown/conflict states.
 const feeHtml = (doc) => {
+  const verifiedFee = independentlyVerifiedFee(doc);
```
J25dunn, line 164: [P1] Preserve monetary-invalid inspector results before promoting the fee. independentlyVerifiedFee recomputes the difference before the existing doc.fee error path. I constructed a raw PSBT whose matching witness/non-witness prevout claims are 2,100,000,000,000,001 sats and whose output is 900 sats. psbtInspectDoc returns fee={known:true,sats:null,error:"amounts exceed Bitcoin's MAX_MONEY"}, but this renderer hides that error and displays 2,099,999,999,999,101 sats as independently verified. Matching transaction bytes do not make an out-of-range amount valid. Please retain the inspector's monetary/conflict errors and add a raw-PSBT regression covering matching claims above MAX_MONEY (and aggregate overflow/range failures).

J25dunn (changes requested):
Reviewed the current head and reproduced a monetary-validation regression using actual PSBT bytes passed through psbtInspectDoc and psbtVizHtml. Please preserve existing inspector errors before promoting a fee to independently verified. Focused tests: 27 passed, 4 failed. Three new assertions expect ungrouped `1000 sats`, while the renderer correctly groups it as `1 000 sats`; the existing column-hint assertion also expects the previous wording. Please update these expectations while retaining their coverage. This review did not include a full interactive browser run.

Hero-Gamer:
Addressed J25dunn's review on `a4fb50a`.

**P1 — inspector monetary errors now win.** `feeHtml` renders `doc.fee.known && doc.fee.sats === null` (MAX_MONEY, u64 overflow, outputs-exceed-inputs) before any locally recomputed independently-verified fee. `independentlyVerifiedFee` also returns null in that case. Matching witness/non-witness bytes no longer hide an out-of-range amount behind `(independently verified)`.

**Tests.** Updated the three `1000 sats` assertions to the grouped `1 000 sats` form, and updated the column-hint expectation to the current wording. Added:
- synthetic MAX_MONEY + matching claims
- raw PSBT through `psbtInspectDoc` + `psbtVizHtml` for `2_100_000_000_000_001` sats / 900-sat output
- raw two-input `u64::MAX` aggregate overflow

Locally: `node --test test/psbt-utxo-verification.test.mjs test/psbt-viz.test.mjs` — 34 passed. Did not run full `test:ci` or a browser-suite pass.

Hero-Gamer:
Thanks @J25dunn — that P1 catch was real, and the grouped-sats / column-hint test mismatches were on us. Appreciate you reproducing it with actual PSBT bytes through `psbtInspectDoc` + `psbtVizHtml` instead of only the synthetic path.

**Review follow-up (on current head `50831bb`):**

- **P1 — inspector monetary errors win.** `feeHtml` now renders `doc.fee.known && doc.fee.sats === null` (MAX_MONEY, u64 overflow, outputs-exceed-inputs) *before* any locally recomputed independently-verified fee. `independentlyVerifiedFee` also returns `null` in that case. Matching witness/non-witness bytes no longer hide an out-of-range amount behind `(independently verified)`.
- **Tests.** The three `1000 sats` assertions now expect the grouped `1 000 sats` form, and the column-hint expectation matches current wording. Added regressions for synthetic MAX_MONEY + matching claims, raw PSBT `2_100_000_000_000_001` / 900-sat output, and raw two-input `u64::MAX` aggregate overflow.

**Rebase:** this branch is now on current `rock` (`813c59a`). Only conflict was `package.json` `test:ci` — rock's list is kept (`journal-wipe-race`, `keymanager-import-render`, `psbt-editor-fuzz`, …) with `test/psbt-utxo-verification.test.mjs` inserted after `psbt-viz.test.mjs`. Diff vs rock is still the same four files.

Mergeable is clean. CI on this head is green, including Test · complete dependency-free suite, Build · WASM from Rust sources, and Test · Firefox integration suite.

src/js/psbt-viz.js
Hero-Gamer: Agreed — matching prevout bytes do not make an out-of-range amount independently verified. On current head, `feeHtml` keeps the inspector error (`doc.fee.known && doc.fee.sats === null`) ahead of any local fee recompute, and `independentlyVerifiedFee` returns `null` in that case. Added the raw MAX_MONEY and `u64` overflow regressions you described. Thanks for catching this.

J25dunn (approved):
ACK on the reviewed revision. Our prior monetary-error blocker is resolved: inspector errors take precedence, and the matching-claim MAX_MONEY and aggregate-overflow regressions pass through the raw PSBT inspection/rendering path. All 34 focused UTXO/visualizer tests pass. Local build passes; current CI is green. Approval covers this follow-up review, not an exhaustive security audit. I did not run a full local browser suite or rebuild Rust locally.

w-s-bitcoin:
Thanks for the substantial work here and for addressing the earlier concerns. We are moving this PR back to draft and placing #191 in the post-v1 backlog.

The direction is valuable, but it introduces a new PSBT verification capability and independently verified fee semantics, rather than a UI/UX refinement during the current feature freeze. The branch also now conflicts with the current rock branch and the recently merged PSBT cost work, so it will need to be rebased and carefully reintegrated if we revive it.

Before that, a focused PSBT/security review should settle the terminology and trust model. In an offline app, we can verify the internal consistency of the transaction and PSBT data supplied by the user, but not chain confirmation or current unspent status, so a Verified label must not overstate what was established. Handling of duplicate or contradictory UTXO fields also needs an explicit review.

I am preserving the PR rather than closing it because this could be a useful EntropyLab feature after v1.

Issue #360, opened 2026-09-05
PSBT editor: tx edits preserve stale signatures; failed edits leave previous PSBT exportable


Rob1Ham:
**Reported via security audit. Medium severity (UX/semantics).**

Two related PSBT-editor behaviors can leave a user holding a misleading file:

1. **Transaction edits preserve stale signatures.** Editing the unsigned transaction (version, locktime, input prevouts/sequences, output amounts/scripts) does not clear existing partial signatures or final scripts, even though any such edit invalidates every BIP143/BIP341 digest those signatures commit to. The edited PSBT still "looks signed" in the flow diagram.

2. **Invalid pending edits leave the previous PSBT fully exportable.** When a pending edit fails validation, the previously-built (still-valid) PSBT remains in the export field rather than being retracted, so the user can export a file that no longer reflects the edits they see in the form.

Both are fail-*visible* at the signature-verification layer but not at the editor's own layer, and a user assembling or finalizing by hand could be misled about what the bytes commit to.

**Suggested fix:** on any unsigned-tx field edit, drop (or loudly mark as invalidated) all existing signatures/final scripts for affected inputs; on a failed edit, retract or disable the export until a clean rebuild.

portlandhodl:
Both halves are fixed by #372 (merged into rock):

- failed edits now retract every export boundary (copy/download/reload/QR) while the displayed fields are stale — commit 03c8ace
- transaction edits drop the signing/finalization fields they invalidate, so the rebuilt PSBT never looks signed with stale material — commit 6943aff

Regression coverage: browser-suite cases drive a valid edit → invalid pending value → recovery, and watch the finalized BIP-174 vector's input flip back to unsigned after an amount edit.

PR #383, opened 2026-09-05
PSBT editor: fuzz input/output editing; fix row-delete double-fire and focus restore

portlandhodl:
## What

Adds a seeded property fuzzer for the PSBT editor's structural editing path, plus the two small fixes it surfaced.

**`test/psbt-editor-fuzz.test.mjs`** (new, wired into `test:ci`): a deterministic PRNG (fixed seed, ~240 walks) drives add/delete input/output, per-field edits (txid, vout, sequence, value, scriptPubKey, version, locktime), and key-value pair add/remove over every rebuildable committed fixture, through the editor's real pipeline — `psbtEditorBuildDoc` → `psbtBuildBytes` → `psbtInspectDoc` — mirroring the handlers' semantics (`mutate()` backup/restore, live field edits that keep typed text on failure, and the `dropSigningPairs` invalidation on transaction changes). Oracles per step:

- **Safe:** every rejection is a clean `Error` (never a `WebAssembly.RuntimeError` crossing the boundary), the WASM instance is re-proven healthy after each rejection, a rejected structural edit leaves the working document byte-identical, a poisoned field blocks structural edits until repaired (or is deleted away), and the walk always reconverges to a building document.
- **Correct:** every accepted build is a byte-exact fixpoint (`build ∘ inspect = id`), added slots decode to the documented defaults with empty maps, deletes keep survivors aligned with their maps (minus the intentionally dropped signing pairs), the unsigned-tx pair regenerates first, and the boundary rejects duplicate outpoints, zero inputs/outputs, amounts past MAX_MONEY, bad hex, duplicate keys, and corrupt documents — all pinned in deterministic guards alongside flavor/volume non-vacuity floors.

**Fix 1 — row delete double-fire:** the pair-delete handler was bound to the bare `.psbted-del` styling class, which the input/output row deletes and the diagram close button also carry. Every row delete double-fired it with an undefined `data-kind`; the resulting `TypeError` surfaced as a spurious error banner and falsely marked the fresh build stale ("The fields do not build right now…"). Now bound to `.psbted-del[data-kind]` only. (Reproduced pre-fix in jsdom, Firefox, and Chrome.)

**Fix 2 — focus restore:** the diagram's sats box and the table row share `data-txout-val`, so `captureFocus()`'s selector resolved to the diagram's copy and typing in the table's sats field moved focus after the first keystroke. Duplicated selectors are now qualified by the edited element's classes.

## Regression coverage

- Browser suite: structural add/delete round-trip asserting no error banner and no stale mark, plus a focus-retention check — verified to **fail pre-fix in both Firefox and Chrome**, pass post-fix.
- Source-guard for the corrected binding in `psbt-editor-tx.test.mjs`.

## Commands run

- `node --test test/psbt-editor-fuzz.test.mjs` (and the editor/PSBT suites)
- `npm run build` — clean
- `npm run test:ci` — 1168 pass, 0 fail
- `npm test` — green except the 5 pre-existing failures from the untracked WIP `test/psbt-network.test.mjs` (not part of this change; it references symbols that don't exist on `rock` — worth deleting or moving to its own branch)
- `npm run test:browser` — green in Firefox and Chrome (Edge absent, skipped by the harness)

`entropylab.html` deliberately not committed (CI rebuilds it).


MrHodlX (changes requested):
BLOCK.

Both UI fixes work: rock double-fires row deletes and steals table focus; this head avoids both. No remaining qualified-selector collision found.

The new fuzzer inaccurately mirrors recovery: it preserves the signing anchor when production clears it. A rejected last-output deletion can preserve last-good bytes, then an accepted transaction edit retains an attached signing field. The rejected deletion also marks the restored, buildable document stale; the fuzzer cannot detect that because it does not drive DOM handlers or track stale state.

Fixed-seed 240 walks, clean-rejection/WASM-health checks, byte-exact fixpoints, and boundary guards pass and are in test:ci. No new dependency, no hand-edited entropylab.html, no test/psbt-network.test.mjs in this PR.

src/js/psbt-editor.js:1033 pair deletion restricted to .psbted-del[data-kind].
src/js/psbt-editor.js:817-819 classes distinguish table and diagram focus targets.
test/psbt-editor-fuzz.test.mjs:186-195 missing anchor reset diverges from production:764-766.
src/js/psbt-editor.js:880-883 rollback restores the document, but not its anchor, and marks it stale.
test/browser-suite.html:2477-2506 delete/error/stale and focus regressions.

Does not close #325, #360.

PR #384, opened 2026-09-06
Detect duplicate key material across multipath descriptor branches (fixes #164)

### LINKED ISSUE #164: Detect duplicate key material across multipath descriptor branches
## Problem

Multipath descriptors can potentially derive the same key material in more than one branch, which can be difficult to spot by inspecting the descriptor alone.

Bitcoin Core has had recent discussion around descriptor sanity checks for duplicate keys and multipath expressions, including cases where duplicate-key checking does not catch collisions outside branch 0.

For a verification-focused tool like EntropyLab, this seems like a useful thing to detect and make explicit.

## Suggestion

When inspecting a descriptor containing multipath expressions, check whether different branches can derive the same concrete public key.

If a collision is detected, show a high-visibility warning, for example:

```text
⚠️ Potential key reuse

The descriptor derives the same public key
in more than one branch.

Branch 0:
  [fpr/84h/0h/0h]xpub.../0/0

Branch 1:
  [fpr/84h/0h/0h]xpub.../1/0

Same derived public key:
  02ab...

This may indicate unintended key reuse or
overlapping wallet branches.
```

For descriptors where no collision is found:

```text
✓ No duplicate derived keys detected
```

The check should be informational/verification-focused rather than preventing the user from using the descriptor.

## Why

This would provide an independent sanity check for descriptors that are difficult to audit manually, particularly as multipath descriptors and more complex wallet policies become more common.

It could also catch issues that are valid descriptor syntax but potentially surprising from a wallet design, privacy, or key-management perspective.

Relevant Bitcoin Core discussion:

* https://github.com/bitcoin/bitcoin/issues/35629
* https://github.com/bitcoin/bitcoin/issues/35985

BIP-388 also has explicit rules around duplicate public keys and overlapping multipath derivations.

## Scope

Initially, this could be limited to detecting duplicate **concrete derived public keys** across the finite multipath branches that EntropyLab can evaluate.

No changes to descriptor generation or wallet behavior are intended.

(just a noob trying to help — please feel free to ignore if me/my clanker do not make sense!)

Hero-Gamer:
## Summary

Adds an informational check for duplicate concrete public keys in output descriptors, including duplicates that occur only after expanding a BIP-389 multipath descriptor.

The check parses and validates the descriptor, materializes every multipath branch, derives each branch at child index `0`, and compares the resulting concrete public keys byte-for-byte.

This is inspection-only. It does not modify descriptors, generate or retain key material, sign transactions, broadcast transactions, or access the network.

Fixes #164.

## Problem

A descriptor can contain apparently distinct key expressions while deriving the same concrete public key more than once.

This is particularly easy to miss with multipath descriptors because a collision can occur in a later materialized branch even when the first branch contains distinct keys.

A representative case is:

```text
wsh(or_i(
  pk(xpub/<0;1>),
  pk(xpub/<2;1>)
))
```

After BIP-389 expansion:

* branch 0 compares `/0` and `/2`, which are distinct;
* branch 1 compares `/1` and `/1`, which collide.

A branch-zero-only check therefore misses the duplicate.

This is the failure mode discussed in Bitcoin Core #35629.

## Approach

The implementation:

1. Parses the descriptor using the existing rust-miniscript descriptor path.
2. Rejects invalid descriptor/multipath expressions before collision analysis.
3. Expands the descriptor with `into_single_descriptors()`.
4. Checks **every materialized branch**, rather than sampling branch 0.
5. Derives every branch at child index `0`.
6. Collects concrete `bitcoin::PublicKey` values from each derived descriptor.
7. Compares the serialized public-key bytes directly.
8. Reports repeated concrete keys with their materialized branch and key position.

No address conversion or address-string comparison is used.

### JS / WASM integration

This PR adds the WASM API and thin JS facade (`descriptorDuplicateCheck()`) plus regression tests; wiring the result into the descriptor inspector's user-facing informational warning (`⚠️ Potential key reuse`) can follow separately as a small UI follow-up.

## Why every branch is checked

The important regression is Bitcoin Core #35629.

That case has a sane first branch but a duplicate in a later branch. Checking only the first expanded descriptor would therefore produce a false negative.

The regression tests explicitly exercise this case and require the duplicate to be found in branch 1.

## Child derivation index

Multipath expansion resolves the finite `<...;...>` dimension, but a descriptor may still contain a wildcard such as `/*`.

Version 1 therefore performs the concrete duplicate comparison at child index `0`.

This is intentionally narrower than attempting to prove that arbitrary ranged derivation paths can never overlap.

## Invalid BIP-389 expressions

Invalid multipath expressions are rejected before duplicate scanning.

Regression coverage includes:

* duplicate values inside a multipath tuple;
* multiple multipath specifiers;
* a multipath specifier inside a key origin.

These are not treated as valid descriptors with "no duplicate detected."

## Key comparison

The comparison is performed on concrete derived public-key bytes.

It does not:

* compare addresses;
* compare descriptor strings;
* compare xpub text;
* convert keys to addresses and compare the resulting strings.

For concrete derived keys, rust-miniscript's `derived_descriptor()` gives EntropyLab `bitcoin::PublicKey` values, including the canonical conversion of Taproot x-only keys. The comparison therefore remains at the public-key level.

## Scope

Included:

* BIP-389 multipath expansion.
* Concrete key derivation at child index `0`.
* Duplicate detection across all materialized branches.
* Informational result data.
* Regression coverage for the later-branch collision case.

Not included:

* Descriptor modification or automatic repair.
* Private-key custody.
* Signing.
* Transaction broadcasting.
* Network access.
* Address-based key comparison.
* Symbolic proof that arbitrary wildcard derivation paths never overlap.

## Tests

The new regression coverage includes:

* Bitcoin Core #35629's later-branch collision;
* explicit duplicate public keys;
* duplicate multisig key material;
* duplicate key material across multipath branches;
* invalid BIP-389 expressions;
* distinct xpub controls;
* distinct derivation-path controls.

The repository's existing Rust/WASM build process should be used to regenerate the WASM bindings; no generated WASM file is hand-edited in this change.

## Verification

The source-level implementation is prepared for the repository's normal verification pipeline.

Please run:

```bash
npm run build:wasm
npm test
npm run verify
```

The CI workflow also rebuilds the WASM artifacts from the pinned Rust sources and tests the freshly generated bindings.

## References

* Bitcoin Core #35629 — duplicate-key sanity checking for multipath descriptors:
  https://github.com/bitcoin/bitcoin/issues/35629
* Bitcoin Core #35985 — multipath descriptor expansion/materialization:
  https://github.com/bitcoin/bitcoin/issues/35985
* Bitcoin Core PR #22838:
  https://github.com/bitcoin/bitcoin/pull/22838
* BIP-389 — Multipath Descriptors:
  https://bips.dev/389/
* BIP-388 — Wallet Policies for Descriptor Wallets:
  https://github.com/bitcoin/bips/blob/master/bip-0388.mediawiki
* rust-miniscript 13.1.0 descriptor API:
  https://docs.rs/miniscript/13.1.0/miniscript/descriptor/enum.Descriptor.html

Fixes #164.


Hero-Gamer:
Not sure why that last test keeps failing: [Test · complete dependency-free suite (ubuntu-latest)](https://github.com/OogaBoogaX/entropylab/actions/runs/34055596434/job/101546873391?pr=384#logs) I tried my best with AI, Terminal and all, still unable to resolve this problem..

Hero-Gamer:
Please note: This is a "Engine only" PR 

"_**wiring** the result into the descriptor inspector's user-facing informational **warning (⚠️ Potential key reuse) can follow separately as a small UI follow-up**._"

src/js/addresses.js
```diff
@@ -169,6 +170,57 @@ export const descriptorDerive = (descriptor, index, network) => {
...
+    }
+  }
+
+  if (!Number.isSafeInteger(childIndex) || childIndex < 0 || childIndex > 2147483647) {
+    throw new Error("Descriptor derivation index must be 0 to 2,147,483,647.");
+  }
+  const bytes = textEncoder.encode(String(descriptor ?? ""));
+  const record = withInput(bytes, (p) =>
+    withOutput(DUPLICATE_DESCRIPTOR_CAP, (out) =>
+      wasm().el_desc_duplicate_check(p, bytes.length, childIndex, out, DUPLICATE_DESCRIPTOR_CAP)
```
J25dunn, line 196: [P2] Exercise this export with the freshly built WASM in CI. The module currently loaded by test:ci does not export el_desc_duplicate_check, so seven new tests fail here with a TypeError before exercising the implementation. The build-wasm job rebuilds the module independently, but its explicit binding-test list does not include test/descriptor-duplicate.test.mjs and its fresh artifact is not supplied to the ordinary suite. Please connect the new tests to the rebuilt artifact and resolve the stale-artifact ordinary test path; the green WASM build alone does not validate this API.

J25dunn (changes requested):
The new duplicate-check tests need to run against freshly built WASM before this can be evaluated for merge. Locally, 1 of 8 focused tests passed and 7 failed with the same missing-export error seen in CI. The separate WASM build job passes, but its explicit test list omits descriptor-duplicate.test.mjs; the ordinary CI suite uses the old checked-in module. Please wire the new suite into fresh-WASM validation and make the ordinary test path consume a matching artifact, following the repository's generated-artifact policy rather than hand-editing generated base64. I did not rebuild Rust locally, so this is not a correctness sign-off on the detector's algorithm. The full CI log also contains separate failures in unchanged PSBT editor fuzz tests; those should be investigated separately rather than attributed to the duplicate detector without evidence.

src/js/addresses.js
Hero-Gamer: Thanks @J25dunn — this was the right catch.

The previous head called `el_desc_duplicate_check` from `descriptorDuplicateCheck()`, but the committed `entropylab-wasm-b64.js` did not export it, and the `build-wasm` job never ran `test/descriptor-duplicate.test.mjs` against the rebuilt module. A green WASM job therefore did not prove this API.

`5d5740f` rebases onto current `rock` and:
- rebuilds `entropylab-wasm` with `npm run build:wasm` so the committed artifact exports `el_desc_duplicate_check`
- keeps `test/descriptor-duplicate.test.mjs` on `test:ci` (ordinary suite now loads a matching module)
- adds that suite to the fresh-WASM job list in `.github/workflows/ci-cd.yml`
- lists the new export in `test/wasm-boundary.test.mjs`

On this head, `Test · complete dependency-free suite` and `Build · WASM from Rust sources` are both green. The PSBT editor-fuzz failures from the earlier log were out of this diff; they are not present on the rebased run.

Appreciate the review.

Hero-Gamer:
Thanks @J25dunn for the review.

Your P2 was correct: the new suite was running against a stale committed WASM module that did not export `el_desc_duplicate_check`, and the fresh-WASM job did not include `test/descriptor-duplicate.test.mjs`.

Addressed in `5d5740f` (rebased onto current `rock`):
- committed artifact rebuilt via `npm run build:wasm` (not hand-edited)
- `test/descriptor-duplicate.test.mjs` wired into both `test:ci` and the `build-wasm` binding-test list
- `el_desc_duplicate_check` added to the wasm-boundary export surface

Current CI on this head is green, including the ordinary suite and the WASM rebuild job. The earlier PSBT editor-fuzz noise was unrelated to this change and is not reproducing here.

J25dunn (approved):
ACK on the reviewed revision for the prior WASM/test-integration blocker. The rebuilt module exports el_desc_duplicate_check; all eight duplicate-check tests pass locally, along with the address and WASM-boundary suites. The fresh-WASM CI job now explicitly includes the duplicate-check suite. Local build passes; current CI is green. This approval is scoped to the reviewed engine change and follow-up validation, not completion of the deferred user-facing warning or exhaustive algorithm auditing. I tested the supplied rebuilt artifact and did not rebuild Rust locally or run a full local browser suite.

w-s-bitcoin:
Thanks for the substantial engine work and the follow-up that corrected the WASM/test integration. We are moving this PR to draft and placing #164 in the post-v1 backlog.

This is a useful direction, but it adds a new descriptor-analysis capability during the current feature freeze. The implementation is also engine-only: nothing in the UI currently calls it, so it does not yet deliver the user-facing warning described by the issue and should not close #164 in its present form. Shipping an otherwise unused WASM API in v1 would add code and review surface without a user-visible benefit.

Before this is revived, the implementation should be rebased onto current rock and reviewed together with its intended UI and warning semantics. In particular, the current check derives remaining wildcards only at child index 0, so any no-duplicate result must not imply that the full ranged descriptor was proven collision-free. The duplicated handwritten BIP-389 pre-validation in JavaScript and Rust should also be reviewed against the parser and supported descriptor grammar, with additional coverage for collisions that appear only at nonzero child indexes and for unsupported or ambiguous cases.

We are preserving the PR rather than closing it because the underlying feature may be worthwhile after v1.

Issue #57, opened 2026-08-29
Mark PSBT input amounts and fees as unverified unless prevouts are cross-checked


Rob1Ham:
## Summary

The PSBT inspector computes input totals and fees solely from `witnessUtxo` values supplied by the PSBT. It does not parse `nonWitnessUtxo` entries or cross-check the claimed amount and script against the referenced previous transaction.

The UI labels the result `Fee (from PSBT fields)`, but the prominent numeric display can still be interpreted as independently verified.

This was identified through static analysis of commit `c4240853fb5bebca583bba1131309035a967e33d`. No runtime proof of concept was executed.

## Severity

- Provisional scanner admission severity: **High, unverified**
- Post-audit triage severity: **Informational**

EntropyLab does not sign or broadcast transactions. This is a misleading-display and defense-in-depth issue, not direct key exposure.

## Affected behavior

- `hodlWitUtxo` reads only input-map type `0x01` (`witnessUtxo`).
- `hodlRenderPsbt` sums those values and subtracts transaction outputs to display the fee.
- Input-map type `0x00` (`nonWitnessUtxo`) is not decoded or used to verify the referenced output.
- If every input has a claimed witness amount, the UI displays a definite numeric fee.

## Impact

A malicious coordinator can supply misleading input amounts and make the inspector display an incorrect fee. The practical effect is contributory: a user could gain false confidence before signing elsewhere, particularly if the external signer also fails to verify amounts.

For SegWit v0, signing commits to the supplied amount, so a contradictory amount generally produces an invalid signature rather than silently changing the spend. Legacy and incomplete PSBT handling should still avoid implying verification.

## Synthetic reproduction plan

1. Build a synthetic PSBT v0 with known previous outputs.
2. Include `witnessUtxo` entries whose claimed amounts differ from the actual referenced outputs.
3. Inspect it in EntropyLab.
4. Observe that the displayed fee is calculated from the supplied claims.
5. Add matching `nonWitnessUtxo` data and confirm the current implementation does not detect the contradiction.

## Recommended remediation

1. Parse `nonWitnessUtxo` when present.
2. Verify its transaction ID against the input outpoint.
3. Verify the selected output amount and script against `witnessUtxo`.
4. Flag contradictions prominently.
5. When only `witnessUtxo` exists, label amounts and fees as unverified PSBT claims rather than verified transaction facts.
6. Suppress a definitive fee display when required inputs cannot be cross-checked.

## Acceptance criteria

[...]

PR #385, opened 2026-09-06
PSBT inspection: expose exact transaction cost and size (fixes #293)

### LINKED ISSUE #293: PSBT inspection: expose exact transaction cost and size
## Problem

EntropyLab already provides detailed offline inspection of PSBT structure, transaction inputs/outputs, PSBT-provided amounts and fees, signing-related fields, and other transaction metadata.

A useful transaction-level layer is still missing: **deterministic cost and size facts for the transaction represented by the PSBT**.

For an already-constructed transaction, users may want to answer:

* How many inputs and outputs does this transaction have?
* How much value is being spent?
* What is the transaction fee?
* What is the fee rate, when it can be determined?
* What is the exact transaction weight?
* What is the exact BIP-141 virtual size once the final transaction is reconstructable?

These are properties of an already-constructed transaction, not wallet or coin-selection functionality. They fit naturally into EntropyLab's existing offline inspection model.

BIP-141 defines weight and vsize in terms of serialized transaction including witness discount. The pinned `rust-bitcoin 0.32.102` API provides `Transaction::weight()` and `Transaction::vsize()` for these metrics.

## Proposed direction

Expose deterministic transaction-level facts alongside existing PSBT inspection data in `PSBT / Nonce`.

First set:

* input count / output count
* input/output amounts
* transaction fee
* fee rate, when determinable
* exact transaction weight, when final transaction is reconstructable
* exact BIP-141 virtual size, when final transaction is reconstructable

Definition: the final transaction is reconstructable when every input has `PSBT_IN_FINAL_SCRIPTSIG` and/or `PSBT_IN_FINAL_SCRIPTWITNESS` present (i.e., the PSBT has passed the Input Finalizer role for all inputs, per BIP-174's Transaction Extractor definition).

Key invariant:

> For an incomplete PSBT, EntropyLab should **not** present a guessed final weight/vsize as exact. If final tx cannot be reconstructed, exact size is unavailable.

This keeps output deterministic and avoids conflating exact facts with estimation.

## Fee provenance

Fee info should preserve distinction between what PSBT claims and what is independently verified.

```text
PSBT-provided UTXO amounts -> calculated fee (based on PSBT claims)
Independently verified prevouts -> calculated fee (backed by verification)
```

A mathematically correct fee is not necessarily an independently verified fee. UI/data model should make that visible, as discussed in:

#57 — PSBT input amounts and fees should be treated as unverified unless prevouts are cross-checked
#191 — independent PSBT UTXO verification

## Existing architecture

```text
PSBT bytes -> Rust / rust-bitcoin parsing -> final transaction data, when available -> exact facts -> weight / vsize / fee / feerate
```
Keeps Bitcoin-specific calculations in existing Rust/WASM layer.

## Scope / non-goals

This proposal does **not** aim to implement:

- UTXO selection or coin-selection algorithm
- transaction construction / signing / broadcasting
- online fee estimation or fee predictions
- guessed exact size for incomplete transactions
- privacy recommendations based on size

UTXO reduction / consolidation is a use case, not the definition of the feature.

## Tests

Deterministic fixtures with independently derived expected values:

- finalized PSBT with known exact weight
- finalized PSBT with known exact vsize
- SegWit tx where witness data materially affects weight/vsize
- incomplete PSBT where exact final weight/vsize is unavailable
- fee and fee-rate calculations where info is present

Expected values should be independently calculated, not produced by duplicating implementation logic.

## Future considerations

Once exact facts exist, they could be reused for inline scenario calculator (e.g., `vsize * user sat/vB`) inside same inspection panel, or side-by-side comparison of multiple wallet-produced PSBT candidates (similar to Core's internal waste comparison in [#22009](https://github.com/bitcoin/bitcoin/pull/22009) and tools like [Coin Smith](https://github.com/abhiraj75/coin-smith-bitcoin) / [PSBT Analyzer](https://github.com/andrewbiv/psbt-analyzer)). This would be a third sub-tab `PSBT Compare` if needed. Not required for this phase.

## References

- [BIP-141 — Segregated Witness](https://github.com/bitcoin/bips/blob/master/bip-0141.mediawiki)
- [BIP-174 — PSBT](https://github.com/bitcoin/bips/blob/master/bip-0174.mediawiki)
- [rust-bitcoin 0.32.102 — Transaction](https://docs.rs/bitcoin/0.32.102/bitcoin/transaction/struct.Transaction.html)

## Open to maintainer guidance
Smallest useful implementation is deterministic inspection layer: expose exact transaction facts, beginning with weight and vsize when final transaction is reconstructable.

Hero-Gamer:
## Summary

Adds deterministic transaction-level facts to existing offline PSBT inspection: input/output counts, summed amounts from PSBT UTXO claims, fee, exact weight/vsize and fee rate when the final transaction is reconstructable.

This is inspection-only. It does not construct, sign, broadcast, estimate fees, or access the network.

Fixes #293.

## Problem

PSBT inspection already exposes structure, inputs/outputs, PSBT-provided amounts and fees, but does not expose:

- exact transaction weight
- exact BIP-141 vsize
- fee rate derived from exact vsize

For an already-constructed transaction, users want deterministic answers:

- How many inputs/outputs?
- How much is being spent?
- What is the fee and fee rate?
- What is the exact weight/vsize once final tx is reconstructable?

These are properties of an already-constructed transaction, not wallet or coin-selection functionality.

BIP-141 defines weight and vsize from serialized transaction including witness discount. Pinned `rust-bitcoin 0.32.102` provides `Transaction::weight()` and `Transaction::vsize()`.

## Approach

- Add small deterministic cost-facts module over existing decoded PSBT inspection document.
- Sum input amounts from existing PSBT UTXO claims (`witness_utxo` / `non_witness_utxo`) using integer satoshi arithmetic.
- Sum output amounts from `unsigned_tx` outputs.
- Derive fee only when every input amount is known and claimed inputs cover outputs.
- Determine finalization: final transaction is reconstructable when every input has `PSBT_IN_FINAL_SCRIPTSIG` and/or `PSBT_IN_FINAL_SCRIPTWITNESS` (BIP-174 Input Finalizer / Transaction Extractor definition).
- Reconstruct final transaction only when `is_finalized` is true; compute exact weight/vsize from reconstructed transaction serialization.
- Compute fee rate as `fee / vsize` only when both are available.
- Incomplete PSBTs do NOT receive guessed weight/vsize — exact size is reported as unavailable. This keeps output deterministic and avoids conflating exact facts with estimation.
- Keep fee provenance at existing PSBT-claim level (`psbt_claim`). Independent UTXO verification remains separate scope per #57 and #191.

No signing, broadcasting, fee estimation, coin selection, network access, or new dependencies.

### JS / WASM integration

This PR adds the Rust/WASM cost-facts API plus thin JS facade and deterministic regression tests. Wiring into `PSBT / Nonce` panel shows:


portlandhodl:
**Blocking correctness issue:** the "exact" size is wrong for witness transactions.

`serializeFinalTx` in `src/js/psbt-cost.js` serializes each witness stack as `varint(count)` followed by the raw item bytes. BIP-144 requires every stack item to be length-prefixed: `varint(count)` then `(varint(len) + item)*`. The missing 1 byte per witness item underreports `weight`/`vsize` and overreports the fee rate for every finalized SegWit/Taproot PSBT (legacy-only is unaffected).

Verified against the pinned rust-bitcoin 0.32.102 using this PR's own test fixture (2 inputs with [71B, 33B] witnesses, 2 P2WPKH outputs): the true values are **832 WU / 208 vB**; the code reports **828 / 207**. The tests pass only because the expected values were computed from the same buggy serializer.

Fix (one line):

```js
const witnesses = hasWitness
  ? finals.map(({ witness }) => concat(varint(witness.length), ...witness.map((item) => concat(varint(item.length), item))))
  : [];
```

then update the encoded expectations in `test/psbt-cost.test.mjs` (828→832, 207→208, fee rate → 10050000/208 = 48317.307692307695, viz string "208 vB · 832 WU") — ideally cross-checked against rust-bitcoin rather than hand arithmetic.

PR #390, opened 2026-09-07
Close the five red-team findings from #389

### LINKED ISSUE #389: Red team rock 2157692: journal wipe race, SP key copies, descriptor path drop, vanity stop, Key Manager HTML
Audited `rock` at `215769243a0d0cab0b3f7c234e3abe344a9cbf5a`. Five findings. No PR opened from this report.

This is an internal red-team writeup. Treat the High items as security work, not drive-by issues. Do not paste attacker markup into a live Journal.

## SEV: Medium — Journal decryption resurrects cleared secrets

Where: `src/js/app.js:12909–12920`, `src/js/app.js:13042`, `src/js/app.js:14012`.

What I did: Started the real asynchronous Journal unlock using an encrypted test notebook, invoked `hodlJournalWipeMem()` before decryption completed, then awaited completion.

What happened: Immediately after wiping, the document was null and the Journal was locked. When decryption finished, the Journal became unlocked again and its seed entry returned. The completion handler unconditionally installs the decrypted keys and document. The lifecycle handler uses this same wipe function.

Why in scope: SECURITY.md promises lifecycle disposal of Journal secrets. This is application state returning after disposal, not inaccessible garbage awaiting collection.

Fix shape: Invalidate pending operations with a session generation counter on clear, lock, and lifecycle transitions, and discard completed results from obsolete generations.

## SEV: Medium — SP cleanup misses writable private-key copies

Where: `src/js/hdkey.js:160–161`; `src/js/app.js:9188`, `src/js/app.js:9205–9210`.

What I did: Instrumented the existing `HDKey.privateKey` getter to retain references to its returned buffers, resolved the published test wallet’s P2TR input, and ran the production sender-key wipe.

What happened: The returned sender scalar was zeroed, but two additional 32-byte private-key buffers remained nonzero. Every getter access creates a copy: the existence check creates one, and the tweak call receives another. Wiping the node cannot wipe either copy.

Why in scope: SECURITY.md claims best-effort byte-buffer cleanup, including sender keys. These are writable application-owned Uint8Arrays.

Fix shape: Read the private-key getter once, retain that buffer, and wipe it in `finally` after verification and scalar construction.

## SEV: High — Multisig descriptor import silently changes the wallet

Where: `src/js/app.js:6520–6527`, `src/js/app.js:6608–6615`, `src/js/app.js:7752`.

What I did: Supplied a valid two-key `wsh(sortedmulti(...))` descriptor whose keys ended in `/0/20/*`, using account xpubs derived from the published abandon…about seed. Compared the original descriptor’s address with the descriptor reconstructed from the imported keys using the shipped WASM reader.

What happened: Import accepted the descriptor but discarded the entire trailing path. The subsequent default receive derivation used `/0/*`. At index zero:

- Original: `bc1qrtyp3wfyejqnv93sq8phywrmwvysqx8avmv49w3c5pjvxfhfce3q8658nn`
- Reconstructed: `bc1qf4jpv99wj36eqez9fzxrzww6sdy97uw5gmgp38t6trqpk5lre8qsv3ttqz`

The displayed/exported wallet therefore differs from the imported wallet, potentially breaking recovery or monitoring.

Why in scope: Incorrect derivations and potential loss of funds are covered by SECURITY.md.

Fix shape: Preserve every descriptor derivation step and branch choice, or reject descriptors the form cannot reproduce exactly.

## SEV: Medium — Stop before worker readiness still sends the seed and starts grinding

Where: `src/js/vanity.js:315–320`, `src/js/vanity.js:348–350`; `src/js/vanity-worker.js:132`.

What I did: Ran the production grinder and worker source under Node’s worker adapter, requesting Stop immediately after Start, before worker initialization finished.

What happened: Message order was init → stop → grind. The readiness handler subsequently transmitted the mnemonic. The worker reset `stopRequested` and completed all 16 test candidates, reporting `stopped: false`.

Why in scope: SECURITY.md describes worker secrets being wiped with the run. Stop currently fails to prevent a not-yet-started run from receiving and retaining them.

Fix shape: Record cancellation before worker readiness, prevent subsequent key dispatch, and terminate or acknowledge cancellation for every worker.

## SEV: High — Imported Key Manager metadata reaches executable HTML

Where: `src/js/keymanager.js:11–21`; `src/js/app.js:1194`, `src/js/app.js:1289`, `src/js/app.js:10574`.

What I did: Encrypted and reopened a crafted `.elkeys` payload through the real Journal APIs, parsed it with `parseKeyVault()`, and passed its address row through the production renderer. The index field contained an SVG with an event-handler attribute.

What happened: Validation accepted the string index, and the renderer emitted the markup and event handler verbatim inside the table. Imported cached results are restored without re-derivation when the user explicitly chooses Use in Key Station. This requires an attacker-controlled encrypted import that the user can unlock, for example a supplied Journal/recovery bundle. Injection at the HTML boundary is confirmed; browser execution was not exercised.

Why in scope: Malicious imported data entering the application’s HTML renderer. The shipped `unsafe-inline` policy provides no event-handler barrier.

Fix shape: Validate imported state against a strict schema, recompute derived results, and render address indexes exclusively as validated numbers or escaped text.

## Validation and limits

The local build passed. Two focused test runs produced 162 passed/4 skipped and 179 passed. The separate PSBT fuzzer produced two failures: its iteration-83 repair incorrectly expected a null prevout to be accepted, then its coverage floor failed. The full-suite attempt also encountered the missing-browser failure and did not complete. Existing local Node/WASM checks continued. The fetched hosted page identified commit `f3b510a`, so it was not treated as the audited rock artifact.

No additional unexpected network primitive was found in the application code. Hosted worker updates and cached operation follow the documented trust model. The exercised SP scan/spend, PSBT export/signature-cleanup, derivation, and wallet-export regressions passed; that does not establish exhaustive correctness.

## Residual risks already admitted by the docs

1. Download provenance and machine integrity remain essential; lookalikes and compromised hosted delivery can replace the calculator.
2. Hosted/Home Screen caching is not an air gap, and connected reopening can update cached code.
3. Strings, browser copies, BigInts, and some dependency memory cannot be reliably erased.
4. Journal files permit offline password guessing; deterministic encryption also reveals identical exports. Weak entropy and brain-wallet inputs remain unsafe.
5. PSBT analysis cannot independently establish prevout truth, complete ownership, or comprehensive signature/nonce safety.

Would I put a real seed in the downloaded HTML on an air-gapped machine after this pass? Not until these five are fixed and checked in a real browser against independent wallets.

portlandhodl:
## What

Verifies and resolves all five findings from the #389 red-team report (audited at `2157692`, which is this PR's base). Each fix is minimal and carries a regression test that **fails on the audited commit and passes here** (verified by running the new tests in a worktree checked out at `2157692`).

### 1. Journal decryption resurrected cleared secrets (Medium)

`hodlJournalUnlock`/`hodlJournalCreate` installed the decrypted notebook unconditionally when the async decryption completed. A `hodlJournalWipeMem()` mid-decrypt (Clear button or the lifecycle auto-clear — both funnel through it) left the journal locked only until the pending operation resolved.

**Fix:** a session generation counter (`hodlJournalGeneration`) bumped by every notebook teardown. Create/unlock capture it before awaiting; on mismatch the completed result is routed to `hodlJournalDiscardOpened()`, which wipes the plaintext document and the verify digest instead of installing them (the CryptoKeys are non-extractable and are simply dropped).

**Regression:** `test/journal-wipe-race.test.mjs` (new) drives the real app.js handlers with the real journal.js crypto: wipe-during-unlock and wipe-during-create leave the journal locked with the decrypted material zeroed; undisturbed unlock/create still work; a stale unlock cannot displace a live session.

### 2. SP cleanup missed writable private-key copies (Medium)

`HDKey.privateKey` returns a **fresh copy** per read. The sender resolver read it twice (existence check, then the tweak/copy for the resolved scalar), and `node.wipePrivateData()` can never reach those copies.

**Fix:** read the getter once into `nodePrivateKey`, use it for the check and the scalar construction, and `fill(0)` it in the `finally` (both branches copy, so the resolved scalar is untouched and still owned by `hodlSpWipeVinKeys`).

**Regression:** `test/sp-send-session.test.mjs` now instruments the getter, retains every returned buffer, and asserts exactly one copy is read and it is zeroed after resolution — for the tweaked P2TR path and the straight-copy P2WPKH path.

### 3. Multisig descriptor import silently changed the wallet (High)

`hodlMsigDescriptorKeyText` validated a trailing path like `/0/20/*` and then **dropped it**, so the imported co-signers reconstructed `…/0/*` — a different wallet than the descriptor names.

**Fix:** after the existing validation, the tail ahead of the wildcard must be exactly the one branch step the form re-derives (`/0/*`, `/1/*`, or `/<0;1>/*`); a BIP45 cosigner step (`45h` origin, cosigner 0) ahead of it stays accepted because the BIP45 compose re-adds it. Everything else — deeper paths, a bare `/*`, a fixed key with no wildcard, branches outside receive/change — is refused with directions instead of rewritten. (The last case required updating one existing test, which now uses `sh(multi(1,key/0/*))`; a fixed-key descriptor names one address the form never derives, so accepting it was the same silent change.)

**Regression:** `test/msig-descriptor-import.test.mjs` gains the issue's exact `/0/20/*` repro plus `/5/*`, `/<0;2>/*`, `/*`, bare-key, and BIP45 cosigner-1/deep-path rejections, and pins that `/0/*`, `/1/*`, `/<0;1>/*` and BIP45 `/0/0/*` still import.

### 4. Stop before worker readiness still sent the seed (Medium)

Message order was init → stop → grind: the readiness handler posted the grind job (carrying the mnemonic/node) even after Stop, and the worker reset `stopRequested` and ground the whole range.

**Fix:** `VanityGrinder.stop()` records the request on the pool; a late `ready` from a worker that was stopped pre-dispatch is answered with termination, the run settles via `onDone({ stopped: true })`, and no key material is ever posted to it.

**Regression:** `test/vanity-wasm.test.mjs` runs the production pool and shipped worker source under node:worker_threads, stops immediately after start, and asserts `stopped: true`, zero work done, and that no worker ever received a `grind` message.

### 5. Imported Key Manager metadata reached executable HTML (High)

A crafted `.elkeys` passes vault validation with an attacker-controlled **string** address index; `hodlAddressTableRows` interpolated `${row.index}` raw into `innerHTML`.

**Fix:** new `hodlAddressIndexHtml()` renders an index only when it is a non-negative safe integer, else escaped text. Applied at the address-table row (the confirmed sink) and the `firstIndex` leads in the HD account and multisig views. All other result-derived values already render through escaping field helpers; the vault parser is intentionally left permissive, with the renderer as the enforced boundary.

**Regression:** `test/keymanager-import-render.test.mjs` (new) round-trips a payload with `<svg onload=…>` as the row index through the real `parseKeyVault` and the real sliced renderer, asserting the payload arrives as escaped text, legitimate indexes render unchanged, and the two lead-index call sites route through the guard.

## Verification

- `npm run build` — clean.
- `npm test` — 1196 passed, 1 skipped, 2 failed. The 2 failures are the PSBT-editor fuzzer's (`psbt-editor-fuzz.test.mjs` "seeded random walks" + coverage floor), which **fail identically on the base commit** `2157692` (verified in a clean worktree) and are the same two the report itself notes; they are unrelated to these findings.
- `npm run verify` — site artifact verified.
- The new/changed regression tests fail on `2157692` with the report's exact symptoms (e.g. vanity `stopped: false`, journal keys/doc resurrecting) and pass on this branch.
- The headless-Firefox journal notepad test ("stamps notes…") is **flaky on the base commit as well** (observed failing about 1-in-7 on `2157692` under load with the identical `::after` background assertion and cascade); the path it exercises (page-tab CSS/focus) is untouched by this PR. The full browser suite passes on this branch.

## Notes

- No locale catalogs touched: the new descriptor-import errors follow the existing plain-`Error` style in that parser, and the translation automation owns `src/locales/`.
- `entropylab.html` is not committed; CI rebuilds and republishes it.
- `test:ci` registers the two new suites (the ci-completeness guard enforces this).

Closes #389.


MrHodlX (changes requested):
BLOCK

At f4051f5, four findings are fixed, but descriptor import still silently changes wallets: wsh(sortedmulti(2,A/0/*,B/1/*)) is accepted and loses both suffixes; reconstruction places both keys on /0/* or both on /1/*. WASM-derived addresses for both reconstructions differ from the original. The deeper-path rejection therefore only partially closes finding 3. No additional secret retention or executable index markup was found in the changed paths. All 62 focused tests pass; the patched suites fail on the pinned base. Build passes; full tests: 1,191 passed, 4 skipped, 1 failure because no browser is installed. No new dependency or hand-edited entropylab.html.

1. Journal — fixed: src/js/app.js:12589 bumps the generation; 12606 discards plaintext; 12932,12954 guard completions. Tests: test/journal-wipe-race.test.mjs:119,155. Independently verified a delayed stale unlock cannot replace a newer live session.
2. SP copies — fixed: src/js/app.js:9213,9237 reads once and wipes in finally. Getter instrumentation passes for P2TR and P2WPKH: test/sp-send-session.test.mjs:187.
3. Descriptor — still broken: src/js/app.js:6547–6549 accepts then discards differing branch choices; 7774 reconstructs a shared branch. Tests at test/msig-descriptor-import.test.mjs:136,163 cover rejection and acceptance, but omit reconstructed-address equality.
4. Vanity — fixed: src/js/vanity.js:367,321 records Stop and terminates late-ready workers without dispatching keys. Zero-work regression: test/vanity-wasm.test.mjs:614.
5. HTML index — fixed: src/js/app.js:1198–1202,1296,7871 escapes table and lead indexes. Tests: test/keymanager-import-render.test.mjs:66,76,91. Other raw row.index occurrences examined are not imported address-result HTML sinks.

DOES NOT CLOSE: #389

MrHodlX (changes requested):
BLOCK

The four other fixes remain intact, but finding 3 still allows a different wallet: wsh(sortedmulti(2,A/<0;1>/*,B/<1;0>/*)) imports successfully because the guard sorts each tuple. BIP-389 requires positional, lockstep expansion: the original expands to (A/0,B/1) and (A/1,B/0), while reconstruction uses (A/0,B/0) and (A/1,B/1). WASM verification confirmed neither reconstructed address matches either original at index zero. The new test explicitly accepts this bypass. The ordinary mixed-branch and /0/20/* cases are rejected, but #389 remains open. The extra fuzzer commit changes only its null-prevout expectations, not any fix. Focused tests pass 65/65; build passes; full suite: 1,194 passed, 4 skipped, 1 missing-browser failure.

Journal: src/js/app.js:12597,12614,12940,12962 — generation bump, plaintext discard, and completion guards remain.
SP: src/js/app.js:9221,9245 — one getter read, wiped in finally.
Descriptor blocker: src/js/app.js:6552,6597,7782 — sorting hides disagreement before shared-branch reconstruction. test/msig-descriptor-import.test.mjs:198–200 incorrectly accepts reversed tuples; the address-equality test at 205–219 omits them.
Vanity: src/js/vanity.js:321,367; test/vanity-wasm.test.mjs:614 — Stop prevents late key dispatch.
HTML: src/js/app.js:1198–1202,1296,7879; test/keymanager-import-render.test.mjs:66 — table and lead indexes remain escaped; no new raw sink introduced.

DOES NOT CLOSE: #389

portlandhodl:
Both blocks addressed.

**Block 1** (mixed branches `/0/*` + `/1/*`) — resolved in `abae68f`: the parser canonicalizes each key's branch choice and refuses the descriptor when co-signer keys disagree, with a rust-miniscript round-trip proving both reconstructions diverged from the original.

**Block 2** (reversed multipath tuples `<0;1>` + `<1;0>`) — resolved in `bd64a53`. The block was correct: BIP-389 expands multipath wildcards positionally (lockstep), so sorting each tuple hid a real pairing disagreement — the original expands to (A/0,B/1) and (A/1,B/0), which no shared-branch reconstruction reaches. The canonicalization now preserves element order; only the one-element multipath still folds into its sole step (`<0>` ≡ `/0`, same expansion).

Changes:
- `src/js/app.js` — branch token is `branches.map(Number).join(";")` in written order (no sort).
- `test/msig-descriptor-import.test.mjs` — the test that incorrectly accepted reversed tuples now pins their rejection; `<0>` + `/0` folding remains pinned as accepted; the rust-miniscript round-trip additionally proves the reversed-pair expansions and both shared-branch reconstructions are four mutually distinct addresses at index 0.
- Negative control: the reversed-tuple rejection test fails on `abae68f` (accepted, "missing expected exception") and passes on `bd64a53`.

Verification: `npm run build` clean; `npm test` 1201 passed / 0 failed / 1 skipped (full suite incl. headless Firefox); new tests fail on the base they patch against.

MrHodlX (changes requested):
BLOCK

At bd64a53, four findings remain fixed and the reversed-tuple bypass is rejected, but finding 3 still has an accept-then-rewrite path. With valid depth-1 [fingerprint/45h]xpub keys, sh(sortedmulti(2,A/0/*,B/0/0/*)) passes the parser and key validators. The BIP45 exception treats both tails as branch 0; reconstruction adds /0/0/* to both keys. Real WASM derivation produces different addresses. Require the BIP45 cosigner step consistently and test reconstructed-address equality. The 65 focused tests pass; build passes; full tests report 1,194 passed, four skipped, and one failure because no browser is installed. No new dependency or committed generated HTML changes.

Journal: src/js/app.js:12600 bumps the generation; :12617 discards plaintext; :12943 and :12965 reject stale completions. Tests: test/journal-wipe-race.test.mjs:119.
SP: src/js/app.js:9224 reads privateKey once; :9248 wipes it in finally. Getter instrumentation: test/sp-send-session.test.mjs:187.
Descriptor blocker: src/js/app.js:6546 collapses the BIP45 step; :6601 accepts matching branch labels; :7785 reconstructs different paths. Reversed tuples now fail at test/msig-descriptor-import.test.mjs:204; equality coverage at :210 misses this BIP45 case.
Vanity: src/js/vanity.js:367 records Stop; :321 prevents subsequent grind dispatch. Test: test/vanity-wasm.test.mjs:614.
Key Manager: src/js/app.js:1198 escapes nonnumeric indexes; table and lead sinks use it at :1202, :1296, and :7882. Tests: test/keymanager-import-render.test.mjs:66 and :91.

DOES NOT CLOSE: #389

MrHodlX (approved):
SHIP

At 79b2fa9, the five reported findings are closed. The BIP45 bypass, mixed branches, reversed tuples, and /0/20/* are refused; supported uniform paths still import. Independent checks using valid BIP45/BIP48 keys, production parsing and reconstruction, and WASM confirmed address equality. No remaining accept-then-rewrite bypass was reproduced in the descriptor-import flow. All 67 focused tests pass; build passes. Full tests: 1,196 passed, four skipped, one failure because no supported browser is installed. No new dependency or committed generated HTML changes.

Journal: src/js/app.js:12606 increments the generation; :12623 wipes discarded results; :12949 and :12971 reject stale completions. Tests: test/journal-wipe-race.test.mjs:119.
SP: src/js/app.js:9230 reads privateKey once; :9254 wipes that buffer in finally. Getter instrumentation: test/sp-send-session.test.mjs:187.
Descriptors: src/js/app.js:6550 requires the BIP45 cosigner step; :6554 rejects unsupported tails; :6561 preserves tuple order; :6606 rejects branch disagreement. Reconstruction: :7791. Tests: test/msig-descriptor-import.test.mjs:136, :163, :181, :198, :239, :256; WASM equality tests at :210 and :272.
Vanity: src/js/vanity.js:367 records Stop; :321 blocks subsequent grind dispatch. Test: test/vanity-wasm.test.mjs:614.
Key Manager: src/js/app.js:1198 escapes nonnumeric indexes; table and lead sinks use it at :1202, :1296, and :7888. Tests: test/keymanager-import-render.test.mjs:66 and :91.

DOES NOT CLOSE: none of the reviewed findings. #389 can close.

PR #391, opened 2026-09-07
Fix remaining mobile derive overflow

### LINKED ISSUE #238: Derive Wallet causes massive GUI disruption on mobile
This appears to be resetting page size abruptly and changes perceived size on a mobile device. 

Instead of the quick snap-back to proper window size, after clicking Derive Wallet the window is large and blank. 

https://github.com/user-attachments/assets/4c0abbbf-d0b7-4e53-95d5-1027ec382c1a

w-s-bitcoin:
Fixes #238.

Follow-up to #291. That PR constrained the visible result content, but the mobile page could still widen when the absolutely positioned accessible label for a hidden WIF was laid out inside the horizontally scrolling address table. Give each private placeholder its own positioning context so that label remains inside the table's scroll/clipping context.

The browser regression now verifies that the WIF label is contained by its private-value placeholder. In a focused Chromium mobile-emulation reproduction at 390 CSS pixels, deriving the same 24-word wallet changed the document width from 390px to 548px before this patch; after the patch it stays 390px before and after derivation.

The first CI run also exposed a stale PSBT-editor fuzz oracle after #373 added null-prevout validation. The application correctly rejects an all-zero txid paired with vout `0xffffffff`, but the oracle still classified every u32 vout as valid. The oracle now mirrors the production rule, with deterministic coverage for creating and repairing a null prevout through either field.

Tests:

- `npm run build && npm test` (1,184 passed, 0 failed, 2 skipped)
- `node --test test/psbt-editor-fuzz.test.mjs` (8 passed)
- `node --test test/ui-defaults.test.mjs` (89 passed)
- `npm run test:browser` (78 Chrome integration checks passed; Firefox and Edge are not installed locally)
- Focused Chrome mobile emulation at 390x844 (document remains 390px before and after Derive)

### DIFF
--- a/src/css/styles.css
+++ b/src/css/styles.css
@@ -626,7 +626,7 @@ details.lab-entropy .wallet-result-messages { margin-top: var(--space-control); 
 .table-private-field-value { display: inline-grid; white-space: nowrap; }
 .private-field-value:not(.secret-placeholder),
 .table-private-field-value:not(.secret-placeholder) { color: var(--blue); }
-.secret-placeholder { display: grid; min-width: 0; max-width: 100%; color: var(--muted); }
+.secret-placeholder { position: relative; display: grid; min-width: 0; max-width: 100%; color: var(--muted); }
 .secret-placeholder > .secret-placeholder-mask,
 .secret-placeholder > .secret-placeholder-message { grid-area: 1 / 1; }
 .secret-placeholder-mask { min-width: 0; max-width: 100%; visibility: hidden; overflow-wrap: anywhere; word-break: break-all; user-select: none; }
--- a/test/browser-suite.html
+++ b/test/browser-suite.html
@@ -3017,6 +3017,11 @@
     assert(!document.getElementById("error").textContent, "Wallet derivation failed");
     const mobileResult = document.querySelector(".key-result");
     mobileResult.style.width = "220px";
+    const tablePrivateValue = mobileResult.querySelector(".table-private-field-value.secret-placeholder");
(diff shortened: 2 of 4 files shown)


MrHodlX (approved):
SHIP

The one-line positioning fix addresses the reported remaining #238 overflow without changing secret rendering or exports. Build and all 97 focused tests pass; CI is green. Local full tests: 1,178 passed, four skipped, one failure because no browser is installed. The exact 390px document-width measurement remains author-reported; the committed regression checks label containment and a narrowed result container.

CSS only: src/css/styles.css:629 adds position: relative to .secret-placeholder; the label remains clipped at :634.
No WIF disclosure: src/js/app.js:1309 renders masks and “Private value hidden,” not WIF bytes. Verified with a secret sentinel. No JavaScript, export, network, dependency, or generated-HTML changes.
Regression: test/browser-suite.html:3022 requires the placeholder to be the label’s positioning parent; :3026 checks container overflow. The test uses a 220px result container, not a 390px document. The CSS assertion at test/ui-defaults.test.mjs:2058 fails on unpatched base.
Redundant oracle rule: test/psbt-editor-fuzz.test.mjs:316 duplicates the null-prevout rejection already present in current rock through #390. The added repair test is at :919. Production consensus behavior is unchanged.
Behind, but mergeable: Current rock is 21ec3b2; a merge simulation completed without conflicts. Scope is exactly four files.

DOES NOT CLOSE: none additionally claimed; no PSBT-consensus closure.

PR #405, opened 2026-09-10
Print or save a single-key card from a derived Key Station key

MrHodlX:
Single-key print card from an already-derived Key Station key. Calculator export, not a generator.

**What it is**
A foldable one-page print of one receive address plus that child's spend secret (WIF or Casascius minikey). Same key + script + address → same card. No new entropy, no Paper Wallet tab.

**Placement**
Print single-key card and Save single-key card sit next to Save watch-only sheet / Save unencrypted private sheet.

**Behavior**
- Private face requires confirm: *This page will show a private key. Cover the private side after printing. Printers keep copies.*
- WIF is torn out of the print DOM on `afterprint`.
- Watch-only xpub: public face only (address + QR + fingerprint + LifeHash). No fold, no fake private side.
- Hidden for multisig, Silent Payments, and vanity grind.
- HD still prints one child, never the seed / passphrase / account xprv.
- Testnet / signet / regtest is a solid black banner so nobody funds the wrong chain.
- LifeHash of the master fingerprint is on both faces so they can be matched after the fold.

Does not print: mnemonic, passphrase, xprv, sibling-script descriptors, BIP-85 children, journal notes.

Generated `entropylab.html` is left to CI.

## Tests
- `node --test test/single-key-card.test.mjs` — 17 passed
- `test/wallet-export.test.mjs`, `test/ui-defaults.test.mjs`, `test/i18n.test.mjs` — passed


src/js/app.js
```diff
@@ -1542,6 +1555,112 @@ function hodlFocusWalletResult() {
...
+    return;
+  }
+  let accept = document.getElementById("single-key-card-confirm-accept"), cancel = document.getElementById("single-key-card-confirm-cancel"), last = document.activeElement;
+  let close = () => {
+    overlay.hidden = true;
+    overlay.classList.remove("is-visible");
+    last?.focus?.({ preventScroll: true });
+  };
+  if (cancel) cancel.onclick = close;
+  if (accept) accept.onclick = () => {
```
J25dunn, line 1611: [P2] Release the private-card callback when the dialog closes. accept.onclick captures onAccept, whose run closure captures the payload containing the WIF/minikey. close() only hides the overlay and restores focus, so cancellation leaves that private payload reachable from a DOM event handler; afterprint clears the print DOM but does not clear this handler either. In an isolated harness, cancelling left accept.onclick attached and invoking it still executed the cancelled callback. This is retention of a live reference, separate from the documented inability to erase immutable strings. Please clear the dialog handlers on every close path and release the captured payload after completion; on acceptance, preserve the callback only long enough to execute the intended operation. Add regressions for cancel/Escape/backdrop close and completion, including the subsequent key-clear lifecycle.

J25dunn (changes requested):
Reviewed the current revision. All 17 single-key-card tests pass, but the confirmation lifecycle leaves a private payload reachable after cancellation/completion. Please address the inline finding and add lifecycle regression coverage. I exercised the actual confirmation function in an isolated DOM harness; no live print-preview or full browser test was performed.

w-s-bitcoin:
Thanks for putting this together. We’re closing this because printing or saving a card containing spend-secret material is outside the scope of EntropyLab v1, particularly during the current feature freeze. Even on an air-gapped computer, sending WIF or minikey material into a print workflow creates risks the app cannot control or reliably clean up: operating-system print spoolers, printer drivers, print-preview or PDF workflows, and the printer’s own storage may retain copies of the job, and many printers are network-connected. Removing sensitive data from the page after printing cannot remove copies already handed to those systems. There is also an unresolved review finding that the confirmation callback can retain a live reference to the WIF or minikey after cancellation or completion, but that should not be read as the only blocker—the larger concern is that we do not want v1 encouraging users to put sensitive key material through a printer workflow. Thanks again for the contribution and for the care put into its warnings and tests.

PR #407, opened 2026-09-10
Chrome Browser Extension

mikeinspace:
Chrome Browser Extension for one-click access. Enable DEV mode to install.
Screenshot 2026-09-10 at 1 25 09 AM


J25dunn (changes requested):
[P2] Make Download EntropyLab deliver a genuinely standalone artifact. The link in entropylab-extension/js/03-entropylab-app.js:2353 (also in the static HTML) still downloads entropylab.html, but the extension's file now loads separate js/*.js files, including the application and WASM bundle. The browser downloads only that HTML, so moving the advertised single file to an air-gapped machine leaves the calculator without its scripts. Please package the original self-contained artifact under a separate filename and point both static and runtime download links at it, or remove/relabel this action so it does not promise a standalone download. Add a smoke test that opens the downloaded file in an otherwise empty directory and verifies the calculator boots offline.

Additional maintainability suggestion: provide a reproducible packaging command from maintained sources and an extension-specific smoke test, so this copied bundle cannot silently diverge from later fixes.

Validation: the manifest requests no permissions or host permissions, and JavaScript syntax checks pass. No live extension installation or browser interaction test was performed; the download finding follows from the link target and its external script dependencies. GitHub rejected the inline comment because the generated JavaScript diff is too large, so the finding is recorded here instead.

w-s-bitcoin:
Thank you for taking the time to put this together and for proposing an easier way to access EntropyLab. We’re closing this because a Chrome extension is a new distribution target and is outside the current v1/feature-freeze scope. EntropyLab’s maintained release architecture is intentionally one self-contained `entropylab.html`; this PR duplicates generated HTML and JavaScript into a second application tree, which would create a separate artifact that could drift from security fixes and normal builds. It also currently has no CI results. J25dunn’s review identified a related functional issue: the extension’s “Download EntropyLab” action downloads only its HTML while that file depends on separate `js/*.js` files, so the advertised download would not boot as a standalone artifact on an air-gapped machine. The extension also lacks a reproducible packaging command and an extension-specific offline smoke test that would detect this kind of divergence. We appreciate the contribution, but we do not want to establish and maintain a second packaging and security boundary in this repository at this stage.

PR #410, opened 2026-09-10
PSBT inspect: report duplicate keys and origin consistency (no combiner)

### LINKED ISSUE #409: PSBT inspect: report duplicate keys and origin consistency
## Problem

EntropyLab's PSBT inspector already keeps every map as `Vec<RawPair>` and still calls rust-bitcoin's `Psbt::deserialize`. Two facts are not surfaced to the user:

1. **Duplicate keys.** BIP-174 forbids the same `type||keydata` twice in one map. Coordinators that merge PSBTs can produce this today (Bitcoin Core #35665: duplicate `PSBT_GLOBAL_XPUB` when merging; #36025: duplicate tap-leaf keys). rust-bitcoin rejects the file, but the compare pane (#253) matches by index and does not name the duplicate, so a user can trust a comparison of two files that Core `decodepsbt` will refuse.
2. **Origin consistency.** A bip32 / tap derivation can name a master fingerprint and path that no global xpub can actually derive. The inspector decodes the origin for display and stops there. A mismatch is a consistency fact about the file, not a chain lookup.

These are offline, file-local checks. They do not decide whether a transaction is safe to sign.

## Proposal

On `psbt_inspect`, attach a two-family sanitize report:

- `duplicateKeys` — scan each map's `Vec<RawPair>` with a first-seen `HashSet` of full serialized keys. Run before `Psbt::deserialize`. Keep the existing `rustBitcoinError` field unchanged.
- `xpubDerivesChild` — for each `PSBT_IN/OUT_BIP32_DERIVATION` and `PSBT_IN/OUT_TAP_BIP32_DERIVATION`, select **applicable** global xpubs (fingerprint match **and** path prefix **and** unhardened suffix), then compare the derived child. 32-byte keydata is compared as x-only. Hardened remaining path, or no applicable xpub, is `incomplete` (NOT CHECKED), not a pass and not a mismatch.

Three-state per #217: `complete` / `problem` / `incomplete`. The editor shows a banner on inspect and a compare footer. Wording is format / origin-consistency, never a safety verdict.

## Out of scope

- Combiner / `joinpsbts` / `combinepsbt`
- Treating one fingerprint mapped to many xpubs as an error (multi-account same fingerprint must pass when the applicable xpub derives)
- Hardware brick narrative, HWI, SeedSigner
- A new workspace tab
- Signing, chain RPC, proprietary-value logging

## Acceptance

- Core #35665 A and B each inspect clean; the hand-combined file reports `duplicate_key` / `PSBT_GLOBAL_XPUB`.
- Same control block twice is a duplicate; two control blocks with one script are not (BIP-371 keys by control block).
- Matching origin derives; mismatch is `problem`; x-only tap derivation matches; hardened gap and missing xpub are `incomplete`.
- Multi-account same fingerprint selects the applicable xpub and passes.
- Existing inspect fixtures stay green. `rustBitcoinError` key unchanged. No new tab. WASM rebuilt via `npm run build:wasm`, not hand-edited.

## References

- BIP-174 duplicate keys; BIP-371 tap leaf keyed by control block
- Bitcoin Core [#35665](https://github.com/bitcoin/bitcoin/pull/35665), [#36025](https://github.com/bitcoin/bitcoin/pull/36025)
- Bitcoin OpTech [#420](https://bitcoinops.org/en/newsletters/2026/08/28/)
- rust-bitcoin PSBT module is v0-only
- EntropyLab #217 three-state, #253 semantic compare

Hero-Gamer:
Fixes #409

Verifier-only. No signing, no chain RPC, connect-src none unchanged.

## Changes

- `psbt-wasm/src/lib.rs` (+ `sanitize.rs`): compute sanitize on inspect. `duplicate_keys` scans `Vec<RawPair>` before rust-bitcoin so a second identical `type||keydata` is a finding instead of a silent overwrite. `xpub_derives_child` selects applicable global xpubs (fingerprint match + path prefix + unhardened suffix) and compares compressed or x-only children. Hardened remaining path is incomplete, not a mismatch.
- `src/js/psbt-editor.js`: inspect banner + compare footer, three-state per #217 (`complete` / `problem` / `incomplete`). Format and origin-consistency facts, not a safety verdict.
- `test/fixtures/psbt/sanitize/*.hex` + `test/psbt-sanitize.test.mjs`

WASM rebuilt with `npm run build:wasm`. Existing inspect fixtures stay green. No new tab.

## Security

Duplicate-key is a format fact. Origin-derive is a consistency fact. Incomplete is visible. Findings do not include pair values (no proprietary payloads). Fingerprints are the 8-hex master fingerprint.

## Test

- Core #35665 A/B each pass individually; combined hand-built file fails `duplicate_keys`.
- Valid two control blocks, one script, passes (BIP-371 keyed by control block).
- Multi-account same fingerprint passes via applicable-candidate selection.
- x-only tap derivation matches; hardened gap and missing xpub are incomplete.

Local: `node --test test/psbt-sanitize.test.mjs test/psbt-wasm.test.mjs test/ci-completeness.test.mjs` — all pass.

## Out of scope

Combiner/join, fingerprint→many-xpubs as error, hardware brick narrative, new tab, HWI, SeedSigner.

## Refs

BIP-371 keyed by control block; Core [#35665](https://github.com/bitcoin/bitcoin/pull/35665), [#36025](https://github.com/bitcoin/bitcoin/pull/36025); OpTech; rust-bitcoin v0-only; #217, #253.


w-s-bitcoin:
Preliminary review notes for @portlandhodl and the author — not a final review verdict. The overall approach is narrow and useful, all GitHub CI checks are green, and the focused local suites (`psbt-sanitize`, `psbt-wasm`, and `ci-completeness`) passed 49/49. I found two cases that appear worth resolving before merge:

1. **Origin analysis has an effectively unbounded multiplicative work path.** `check_one_derivation` calls `applicable`, which scans every global xpub for every derivation, then attempts each applicable child derivation until one matches. I built a 25,872-byte PSBT containing 200 repeated global xpub records and 200 repeated mismatching input derivations. Inspection took approximately **5,900 ms on this PR versus 16 ms on current `rock`**. The existing limits allow substantially larger inputs, so a small hostile file can synchronously freeze the inspector for much longer. Suggested direction: deduplicate/index candidates and enforce a strict derivation-work budget; if the budget is exhausted, report this family as `incomplete` rather than continuing. Please add a regression that establishes a bounded behavior.

2. **Key encoding is inferred from observed length rather than the derivation field type.** `pubkeys_match` accepts compressed, uncompressed, or 32-byte x-only forms for every derivation record. I modified the matching fixture in two ways:
   - `PSBT_IN_BIP32_DERIVATION` with the compressed prefix removed, leaving the matching 32-byte x coordinate;
   - `PSBT_IN_TAP_BIP32_DERIVATION` with the matching compressed prefix added, producing a 33-byte key.

   rust-bitcoin correctly rejected them as `invalid secp256k1 public key` and `invalid xonly public key`, respectively, but both returned `sanitize.xpubDerivesChild.state === "complete"` with no findings, so the new banner independently says the origin check completed. Suggested direction: validate keydata shape according to the PSBT record type before comparison—legacy BIP32 versus Taproot x-only—and classify malformed records as `incomplete` or `problem`. Add both malformed-shape regressions.

The duplicate-key vectors, applicable-xpub selection, hardened-gap handling, and existing-file compatibility looked sensible in this pass. Leaving the final technical review and severity assessment to @portlandhodl.

Hero-Gamer:
Inspired by Evan's PR - asked my clanker to do a quick mock-up, not 100% but gets the idea across I think. Helped me a lot to understand.

I'll try to do this going forward for my PRs.

To be clear - once merged, only hand right side information is displayed, left hand side is what it currently (pre-merge) looks live.

This should bring us in line with Core if I am not wrong!

image

BTW thank you Portland for fixing reviewer's two issues.

PR #411, opened 2026-09-10
Add Lightning node key tool (aezeed/LND and BIP-39/LDK)

kaloudis:
## Summary

Screenshot 2026-09-10 at 6 55 09 PM

Adds a **Lightning** workspace tab (between PSBT and Journal) that derives the node identity public key a seed phrase corresponds to:

- **aezeed (LND)**: the 24-word cipher seed deciphers in WASM (scrypt N=2^15/r=8/p=1, AEZ v5, CRC-32C) and its 16-byte entropy seeds BIP32 directly, per LND; the node key sits at `m/1017'/coinType'/6'/0/0` (coin type from a mainnet/testnet select). A wrong passphrase is a named error (AEZ authentication failure), distinct from a mistyped word (checksum mismatch) and from an unsupported cipher seed version.
- **BIP-39 (LDK, ldk-node convention)**: standard BIP39 seed, BIP32 master key, its raw private key re-seeds a second BIP32 tree, node secret at `m/0'` (network-independent), per ldk-node's builder and rust-lightning's `KeysManager`.

Decoded aezeed internals (cipher seed version, wallet birthday as day count and UTC date, entropy, salt) and the BIP32 root xprv (the `chantools showrootkey` equivalent, importable into Sparrow) sit behind a reveal toggle. Decoding only: the tab never creates seeds, in keeping with the entropy policy.

## Implementation notes

- **Vendored AEZ v5** (`entropylab-wasm/src/aez/`, from the `zears` crate 0.2.1, MIT, provenance and license notice in `mod.rs`): the crate declares `cpufeatures` unconditionally, which does not compile for wasm32-unknown-unknown, so the cipher core is vendored with the x86-64 AES-NI and nightly-simd paths stripped and its reference test vectors kept. The encipher path exists only under `#[cfg(test)]` so the vectors prove both directions of the port.
- **scrypt** is the pinned RustCrypto crate (`=0.12.0`, default features off). New exports `el_aezeed_decipher` and `el_scrypt` take caller-fixed KDF parameters like `el_pbkdf2_hmac_sha512`; LND's published test vectors were generated with weakened scrypt (n=16), which is exactly why the parameters cross the boundary.
- New JS facades `src/js/aezeed.js` (11-bit word packing, decode, error taxonomy) and `src/js/lightning.js` (DOM-free `deriveLndNode`/`deriveLdkNode` plus the card wiring). Secret hygiene follows the existing conventions: scrypt key and plaintext wiped in WASM, JS temporaries wiped in `finally`, new fields in the lifecycle sweep, secrets reveal-gated. Known residuals (scrypt's internal 32 MiB working buffer, the AEZ key schedule) are documented in the lib.rs header.
- Locale catalogs changed only via `npm run i18n:sync` (dead-entry pruning; translations arrive via the post-merge workflow).

## Testing

- `cargo test`: LND's published cipherseed vectors (weakened params), a full-strength characterization vector from guggero/cryptography-toolkit's e2e suite, the AEZ reference vector suites (extract/E/hash/encrypt/decrypt), CRC-32C check value, scrypt RFC 7914 vector, error taxonomy.
- `test/aezeed-wasm.test.mjs` (registered in `test:ci` and the CI build-wasm job): the same vectors against the artifact, scrypt differential vs @noble/hashes, 11-bit packer round trips, taxonomy across the boundary.
- `test/lightning.test.mjs`: LND and LDK derivations differential vs @scure/bip32 + @scure/bip39 (trezor mnemonics, both passphrases, both coin types), a guard that the two-stage LDK derivation is not the plain `m/0'` of the seed, and wiring assertions.
- Browser suite: a new interaction test drives the tab end to end (aezeed derive, reveal gate, wrong-passphrase error, format switch, LDK derive, wipe) and the tab-strip tests now cover eight tools.
- `npm run ci` green (1232 tests, reproducible build verified).
- Manually verified against real nodes: mainnet LND (aezeed), LDK with a 12-word and a 24-word BIP-39 seed.

## Follow-up ideas

- **Message signing**: sign/verify with the derived node identity key (LND `signmessage` compatible, zbase32 over the double-SHA256 of `Lightning Signed Message:` plus the message), so ownership of a node pubkey can be proven from cold storage.
- **LNURL-auth**: derive the LUD-05 `linkingKey` (hashingKey at `m/138'/0`, per-domain linking keys from HMAC over the domain) and sign the k1 challenge, enabling air-gapped LNURL-auth login proofs.
- **Feature parity with guggero's cryptography-toolkit aezeed page**: the pieces of https://guggero.github.io/cryptography-toolkit/#!/aezeed that fit this repo's policy, such as re-enciphering an existing cipher seed under a new passphrase (a deterministic transformation of user input, LND's changepassword equivalent) and testnet extended-key display for the decoded root. Seed generation stays out per the entropy policy.
- **aezeed encipher from user-supplied entropy**: build a 24-word aezeed cipher seed from entropy the user brings (dice rolls, coin flips, hex), so LND users do not have to rely on the node's internal seed generation. This is the calculator model applied to aezeed: 16 bytes of user entropy plus a user-supplied 5-byte salt, birthday, and passphrase encipher deterministically (same inputs, same words). The vendored AEZ encipher path already exists under #[cfg(test)] and would move into the export surface.

## Attribution

The vendored AEZ module keeps its MIT notice (Copyright 2025 Daniel Schadt, codeberg.org/dunj3/zears); unlike the rest of the repo it is not public domain. Test vectors are transcribed from lightningnetwork/lnd (MIT) and guggero/cryptography-toolkit (MIT) with source comments.


MrHodlX:
Freeze exception is this PR only. Decode-only Lightning tab. No generate, no signmessage, no LNURL, no encipher-from-dice.
Independent pass on ad57e04 against LND cipherseed/keychain and ldk-node KeysManager. Derivations are right. Encipher is #[cfg(test)] only and not in the artifact. Follow-ups from the body are not in the diff.

YES AFTER the three items below. Rebase onto current rock, drop entropylab.html, src/js/entropylab-wasm-b64.js, and the locale files from the PR. CI rebuilds the artifacts; translate workflow owns catalogs.
Blockers

1. High — MIT notice does not ship. Vendored AEZ (entropylab-wasm/src/aez/, zears 0.2.1) is MIT. mod.rs says the built artifact must retain the notice. Grep for Schadt/zears is 0 hits in entropylab.html, the b64 header, and the stripped wasm. The only surviving MIT block in the HTML is LifeHash. CONTRIBUTING §5 still names LifeHash as the sole exception; crate license is still Unlicense.
Fix: put the zears notice on the build-wasm.mjs blurb the same way LifeHash is preserved; confirm the inliner keeps it in entropylab.html; CONTRIBUTING exception line; license = "Unlicense AND MIT". Maintainer sign-off on the exception is the freeze exception.

2. Med — el_scrypt is not bounded. el_scrypt / el_aezeed_decipher check out_len and 1 ≤ log_n ≤ 20. r and p go straight into scrypt::Params::new. At log_n=20, r=8 the working buffer is 1 GiB; it grows with r until wasm32 dies. UI path is fixed at 15/8/1 (aezeed.js:27-29) and never calls el_scrypt — only tests do. The header’s “32 MiB” / “narrow boundary” is false.
Fix: el_aezeed_decipher accepts only (r,p)=(8,1) and log_n ∈ {4,15}. Cap el_scrypt at 128·r·2^log_n ≤ 32 MiB and p ≤ 16, or drop the export and keep the RFC differential in cargo test. Boundary test: log_n=16 and r=9 return -1.
9. Med — scrypt B/V survive Clear and pagehide. After aezeedDecode → deriveLndNode → JS wipe, linear memory still holds V[0] = B = PBKDF2-HMAC-SHA256(passphrase, salt, 1, 1024). Clear / pagehide never touch WASM heap. scrypt 0.12 has no zeroize; b/v are plain vec!. Anyone who can read page memory can test passphrase guesses at one PBKDF2 iteration instead of scrypt $N=2^{15}$. SECURITY.md says the bindings erase HMAC/PBKDF2 blocks. The lib.rs header calls this buffer short-lived. It isn’t.
Fix: after every scrypt::scrypt, success or failure, allocate vec![0u8; 128·r·2^log_n] and vec![0u8; 128·r·p], wipe_bytes both. Node test: decode at production params, then assert pbkdf2(pass, salt, 1, 1024)[0..32] is absent from the heap. Once (2) caps r/p, the scrub is ≤ 32 MiB. Until this lands: close the tab, don’t Clear.
Non-blocking, same PR if cheap

3. Gate internal version to 0|1 (IsKnownVersion). Don’t print a node key for v≥2 on the v0/v1 path. Label the field “internal (key-derivation) version”, not “Cipher seed version”.
4. Format/Network change must hodlLnWipeMem(); hodlLnRender();. Stale mainnet pubkey next to a Testnet select is how people fail lncli getinfo. Help text: coin type 1' is testnet/signet/regtest. Browser test currently switches format only after a wrong-passphrase wipe, so it never covers revealed result → switch.
5. README/PR say version + birthday are behind reveal. Only entropy/salt/xprv are. Move them or fix the docs.
7. SECURITY.md needs the new KDF, the MIT cipher, the 32 MiB residual, and the scrub. Same PR per CONTRIBUTING §5.

3–5 and 7 can land with 1/2/9. Don’t hold the tab for i18n of the #ln-out template; post-merge translate workflow.
What I checked
LND n=16 vectors and the guggero $2^{15}$ vector recompute. Reverse: 3 fresh seeds × coin types 0/1, internal v0/v1, birthday 65535, non-ASCII passphrase; wrong passphrase → passphrase; literal "aezeed" == empty. LDK 12/24-word Trezor set × "" / TREZOR / pässwörd = 12/12 two-stage vs @scure; single-stage m/0' differs in all 12. Footgun guard at test/lightning.test.mjs:84-90 is real. Vendored src/aez/ matches zears 0.2.1 minus the stated AES-NI/simd strip. No storage, no fetch. 8 tabs. Rebase conflict is only entropylab.html.
Not run here: cargo test / build:wasm (CI), browser suite, chantools / live lncli getinfo, NFD vs NFC on the LDK passphrase, bfcache after pagehide.
Would I type a real aezeed into the CI-rebuilt HTML on an air-gapped machine after 1, 2, and 9? Yes. Set Network first. Re-derive after any select change. Check the pubkey against lncli getinfo before the xprv.

PR #420, opened 2026-09-11
Confirm before deriving from below-recommended entropy (#416)

### LINKED ISSUE #416: Low Entropy warning and confirm UI on Derive Key action
We currently activate the Derive Key button after any amount of entropy is supplied by the user. We show an estimated amount of entropy and recommendations but a careless or less astute user may miss the hints. I propose we dynamically add a second step to the key creation process when less than the recommended amount of entropy is supplied. 
 
- User starts interacting with the Key Station to create a new key
- User clicks Derive Key action button when it is enabled (currently happens as soon as ANY entropy is provided)
- if user has provided at least the recommended amount of entropy key is generated (current functionality)
- if provided entropy is below recommendation, a modal appears "You have provided only X entropy..."
- "Add More Entropy" button cancels key creation and user can continue entering more
- "I Understand, Proceed" button creates the key 

We could also include a checkbox "Dont show this again" that when checked would bypass this functionality for that user.

hotpixelgroup:
Closes #416.

## What and why

The Key Station's Derive Key button enables as soon as **any** entropy is supplied, and the per-source "below the recommended amount" hints are easy to miss. This adds the second step the issue asks for when Derive Key is pressed with less entropy than recommended:

- A modal states the estimate: *"You have provided only about X bits of entropy (recommended: Y bits for a N-word seed)."* plus the source detail (e.g. "6 of 99 recommended rolls").
- **Add More Entropy** (also Escape / backdrop click) cancels and keeps the entered rolls/cards so the user can continue.
- **I Understand, Proceed** derives the key.
- **"Don't show this warning again"** remembers the bypass in `localStorage` (same site-settings store as the theme and beta banner). When storage is unavailable (file://, private modes) persistence no-ops and the warning always shows — the safe direction for a wallet tool. Focus lands on the safe choice (Add More Entropy) and is restored on close; the card is a `role="dialog"` `aria-modal` overlay following the existing overlay paradigms.

## Scope decisions

The gate applies only to the two fresh-entropy sources that can actually derive below their recommendation:

- **Hashed dice rolls** (COLDCARD base-10 and Keystone/Ian Coleman methods): warns when `rolls < config.hashRolls` (99 for 24 words, 50 for 12, …), estimate `rolls × log2(6)`.
- **Hashed card transcripts**: warns when below the recommended deal (52+6 for 24 words), without-replacement deal entropy.

Deliberately not gated (documented in a comment on the predicate): D++/BitBox dice and direct cards only enable Derive once their full construction is entered; number bases require the exact digit count; seed-phrase and private-key modes import *existing* material with no "recommended amount"; the brain wallet keeps its own explicit acknowledgement; the multisig station derives from co-signer xpubs, not user entropy. The vanity "Update key" write-back path re-derives an already-created key and intentionally bypasses the modal.

## Notes

- New English strings are keyed at the call sites only; `src/locales/` is untouched — translations are left to the post-merge translation automation (missing strings fall back to English).
- The dismissal persists across app versions (not version-keyed like the beta disclaimer), matching the issue's "bypass this functionality for that user" wording.
- Built artifact `entropylab.html` is not committed, per convention — CI rebuilds it.

## Commands run

- `npm run build` — pass (`entropylab.html` 6,500,614 bytes)
- `npm test` — 1280 tests, 1274 pass, 0 fail, 6 skipped (engine skips)
- `npm run test:browser` — Chrome/Chromium: 84/84 checks passed, including the new "deriving below the recommended entropy asks for confirmation first" (Firefox/Edge not installed on this machine; skipped by the harness)

New tests: `test/low-entropy-confirm.test.mjs` (predicate boundaries incl. 12-word thresholds, storage failure modes, dialog structure, handler wiring invariant; registered in `test:ci`) plus the browser-suite test driving the full appear / cancel / proceed / persist / bypass flow.


w-s-bitcoin:
Thanks for the thoughtful implementation and especially the thorough predicate, storage-failure, and browser-flow coverage. We are closing this PR because the recent Key Station design work in #424 has substantially changed the shared control and UI vocabulary, and this branch now conflicts with the current rock branch across app.js, styles.css, and the browser suite. Adapting it would require enough UI redesign that keeping this implementation open is more likely to create conflicting work for @timechainb than to simplify the eventual change.

Issue #416 will remain open. The underlying low-entropy confirmation is still a useful v1 safety/UX improvement, and the threshold logic and tests here can be referenced when it is implemented within the finalized design system. At that point we should also decide whether a permanent localStorage dismissal is appropriate for a security warning and ensure the modal fully contains keyboard focus or makes the background inert.

Thank you again for the work and clear PR description.

PR #422, opened 2026-09-11
Fixes #418 only — BIP-341 taproot commitment verification

Hero-Gamer:
Fixes #418 — BIP-341 taproot commitment verification only.

**Scope:** This PR is taproot-only. Cost-facts fix for #293 was accidentally bundled in earlier commit and has been removed — it lives in #388 which is the correct fix for #293.

**What this implements:**
- `verify_taproot_commitment` now takes `leaf_script` from `PSBT_IN_TAP_LEAF_SCRIPT` and computes `TapLeafHash`
- Sorted `TapBranchHash::from_nodes()` fold over control block siblings
- Checks `reconstructed_root == merkle_root`, parity, internal_key, `m<=128`, `tweak < n`
- Key-path-only case `m==0 && merkle_root==None` via `TapTweakHash::from_key_and_tweak(P, None)`
- Wired into existing `psbt_inspect()` as `taproot_commitment: {status}` — no new WASM export

**Verification:**
- `None` for missing fields (incomplete), `Some(false)` for invalid commitment, `Some(true)` for valid
- Adversarial tests a-f (m=129, leaf_version 0xC1, parity flip, zero key, zero tweak, garbage siblings)
- BIP-341 wallet vectors both parities, 200×128 siblings perf <100ms
- `cargo test -p psbt-wasm` + `node --test`

Closes #418

Issue #418, opened 2026-09-11
Add Taproot commitment verification to offline transaction inspection


Hero-Gamer:
## TL;DR

Add a verification-only Taproot commitment section to EntropyLab’s existing PSBT/raw transaction inspection workflow.

For supplied Taproot data, show and verify:

```text
TapLeaf
→ Merkle branch
→ Merkle root
→ TapTweak
→ output key
→ P2TR scriptPubKey/address
```

For a revealed script path, validate the supplied tapscript and control block, including control-block length, leaf version, internal key, Merkle proof, output-key parity, and final output-key match.

A future Part 2 can build on the shared verifier for descriptor cross-checking, BIP371 PSBT consistency, Miniscript policy analysis, and related Taproot artifacts.

## Problem

Taproot script-path artifacts contain several linked commitments that are difficult to verify manually:

```text
tapscript
→ TapLeaf hash
→ Merkle branch
→ Merkle root
→ TapTweak
→ output key
→ P2TR scriptPubKey/address
```

Correctness depends on details such as:

- the TapLeaf tagged hash;
- the leaf version;
- control-block length;
- the internal x-only public key;
- lexicographic TapBranch ordering;
- sibling-hash ordering;
- the TapTweak tagged hash;
- output-key parity;
- the final witness-program/scriptPubKey comparison.

A single altered byte can produce a different output key or make a revealed script path unrelated to the supplied output.

Bitcoin Core and libraries implement the underlying rules, but users generally do not receive a compact, offline explanation of the entire intermediate commitment chain. A focused reconstruction view would help users independently inspect supplied wallet, transaction, recovery, and script-path artifacts.

Related external discussion:

- Delving Bitcoin: RootScope deterministic reconstruction of Taproot script-path commitments  
  [https://delvingbitcoin.org/t/rootscope-deterministic-reconstruction-of-taproot-script-path-commitments/2322](https://delvingbitcoin.org/t/rootscope-deterministic-reconstruction-of-taproot-script-path-commitments/2322)
- Bitcoin Core Taproot functional tests  
  [https://github.com/bitcoin/bitcoin/blob/master/test/functional/feature_taproot.py](https://github.com/bitcoin/bitcoin/blob/master/test/functional/feature_taproot.py)
- Bitcoin Core wallet Taproot tests  
[...]

PR #426, opened 2026-09-12
docs: require accepted and rejected validation tests

Hero-Gamer:
## Why

Bitcoin Core's contributing guidelines require:
> Have unit tests, functional tests, and fuzz tests, where appropriate; Where bugs are fixed, where possible, there should be unit tests demonstrating the bug and also proving the fix.

We don't have that explicitly documented for validation code, and it cost us time and money (well.. Mr Hodl's.. 🥲 - I don't like to see money getting wasted.. ) to retroactively red-team it constantly.

## Case study: #418 / #422

Issue #418 listed 5 negative cases, but initial AI implementation in #422 tested only valid cases. Result: 6 BLOCKING bugs shipped in the first implementation PR submission (version check used `!= 0xc0` instead of evenness, zero key, invalid tweak, bad length, etc). Caught only by post-push hunt and fixed in b5626d6.

In Bitcoin Core, this would have been caught because `feature_taproot.py` requires both valid and invalid vectors. We should adopt same practice.

## What this PR does

Documents Core's practice for our repo:

> For any PR touching verification, consensus, cryptography, script, or PSBT parsing: include both valid and invalid vectors. Invalid must fail before fix and be rejected after.

This is inspired by Core's guidelines (not verbatim quote, but same intent). Adds to AGENTS.md + creates.github/pull_request_template.md.

image

Reference: https://github.com/bitcoin/bitcoin/blob/master/CONTRIBUTING.md

Docs-only.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -81,3 +81,17 @@ Guidelines for AI coding agents.
   it, rather than sweeping the suite, so the design diff stays reviewable and
   the count comes down as the work moves through each surface.
 - Before finishing, run `npm run build && npm test` and make sure they pass.
+
+### Testing - valid AND invalid (borrowed from Bitcoin Core)
+
+Following Bitcoin Core's contribution guidelines: https://github.com/bitcoin/bitcoin/blob/master/CONTRIBUTING.md
+
+For any PR touching verification, consensus, cryptography, script, or PSBT parsing:
+
+1. Every PR must include tests
+2. For validation logic, include BOTH valid and invalid vectors
+3. Invalid vectors must fail before your fix and be correctly rejected after
+
+This is not AI-specific - it's the same rule Bitcoin Core uses for `feature_taproot.py`.
+
+If your issue lists "Negative cases", copy them into tests first.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -82,16 +82,16 @@ Guidelines for AI coding agents.
   the count comes down as the work moves through each surface.
 - Before finishing, run `npm run build && npm test` and make sure they pass.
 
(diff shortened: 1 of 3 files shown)


AGENTS.md
```diff
@@ -81,3 +81,17 @@ Guidelines for AI coding agents.
   it, rather than sweeping the suite, so the design diff stays reviewable and
   the count comes down as the work moves through each surface.
 - Before finishing, run `npm run build && npm test` and make sure they pass.
+
+### Testing - valid AND invalid (borrowed from Bitcoin Core)
+
+Following Bitcoin Core's contribution guidelines: https://github.com/bitcoin/bitcoin/blob/master/CONTRIBUTING.md
```
remcoros: this might make the model do a webfetch and leak bitcoin cores own contributing guide into the context. I think it's better to not have arbitrary links in AI instructions, just state hard rules.

w-s-bitcoin:
Updated in 06a3c50 to address the review and keep the policy self-contained:

- removed the external attribution and link from AGENTS.md;
- scoped accepted/rejected coverage to security-sensitive validation and parsing changes;
- made pre-fix regression coverage conditional on practicality;
- allowed documentation-only and presentation-only changes to remain N/A;
- added the same contributor-facing rule to CONTRIBUTING.md;
- made the pull-request-template checkbox explicitly conditional.

Verification: npm run build passed. npm test passed 1,277 tests with 0 failures and 2 environment skips (Firefox and Edge unavailable); Chrome completed all 86 browser checks. The initial sandboxed run had eight EPERM failures because loopback binding was blocked, and all passed when rerun with loopback access.

Issue #423, opened 2026-09-12
Red team: Kimi K3 full-repo pass on rock @ e225ef3


MrHodlX:
Local Kimi Code, model moonshotai/kimi-k3, branch rock, SHA e225ef324a9a2a56867c1f195db94ee0b4fb0844.

# EntropyLab rock red team — full-repo pass

**SHA:** `e225ef324a9a2a56867c1f195db94ee0b4fb0844` (rock; the sweep started at `f59e8a3` and two artifact commits landed mid-review — every finding below was re-verified at `e225ef3`)
**Docs read:** SECURITY.md, README.md, AGENTS.md (current revisions), the `813c59a..e225ef3` diffs of both, and #399's prior pass. Method: 12 parallel module auditors + first-hand verification of every claim below; nothing here is second-hand.
**Scope:** all of `src/js/` (50 modules incl. 14.9k-line `app.js`), all three Rust crates, build/CI workflows, service worker, i18n pipeline.

**Would I put a real seed in the downloaded HTML on an air-gapped machine after this pass?** The repo now answers for me: README opens with "DO NOT USE WITH ACTUAL BITCOIN IN ANY WAY — TESTNET | SIGNET | REGTEST USE ONLY," so no, and that banner is now the load-bearing control. If the banner ever comes down: the crypto core is in good shape (facades differential-match noble/scure; BIP-85/39/32/352 vectors pass; no export generates randomness; no new High anywhere), but the gaps below are exactly the class that bites real-value users — the lifecycle wipe misses rendered seed words (F3), a stopped vanity grind lies about coverage (F2), and the WASM supply chain has one structural lie (F1) plus a docs lie (F6). Testnet-only, I'd use it today.

## Already in the model (not re-flagged)

Empty journal password = no access protection; WASM linear memory never shrinks and tab-close is the only guaranteed erasure; strings/BigInts best-effort; PSBT prevouts unverifiable; nonce-history file is correlation-sensitive; hosted cache ≠ air gap; attestation/provenance limits; weak entropy. Confirmed still admitted and accurate.

### Medium

**F1 — WASM supply chain: the tested build is discarded; an untested second rebuild is committed; the shipped HTML embeds neither**
[...]

Issue #425, opened 2026-09-12
GPT-6 red-team audit (2737bbb)


MrHodlX:
Independent GPT-6 red-team of `rock` at `2737bbb536f095b1f1c2d5ee7003f03ed68cea00`. Not a patch. Calculator, not a generator.

A Fable 5.1 pass on the **same SHA** is still pending. Do not treat GPT-6's "I would not type a real seed" as the project conclusion.

## Maintainer triage (read this first)

| # | GPT-6 claim | Keep? |
|---|---|---|
| 1 | Clear / pagehide does not cancel in-flight derive/import | **Yes.** Journal already has `hodlJournalGeneration` (#389). `hodlWipeActiveKey` and the pagehide wipe do not set `hodlActiveDerivation.cancelled`. The await can write the wallet back onto a cleared key. Fix: generation token on wipe + pagehide + lock; reject stale results; cancel the active derivation control. |
| 2 | PSBT inspect leaves copies in the live WASM heap | Residual, already in SECURITY.md (best-effort wipe, linear memory never shrinks). Recreate the PSBT module on wipe if we want a harder bound. Not a 0.2.0 blocker. |
| 3 | `.elkeys` shows attacker-cached address/WIF without rederive | **Yes, authenticity not XSS.** #389 closed the HTML injection. Parser still copies `result` wholesale; "Use in Key Station" restores it. Untrusted `.elkeys` can display a fake spend secret. Until verified, imported cache is watch-only and must not be labeled spendable. |
| 4 | BIP32 `I_L == 0` retries instead of child = parent | Spec nit from #359. Comment in `el_hd_ckd_priv` is wrong (BIP32 allows `I_L == 0`). Probability ~2^-256. Leave it. |
| 5 | Committed HTML stamps `d5d6a46` not HEAD `2737bbb` | False alarm. Normal two-commit CI (merge, then Rebuild site artifact). `SHA256SUMS.txt` matches the HTML bytes. |

**Air-gap `file://` dice → address:** still the intended path. Don't import a stranger's `.elkeys` until #3 is fixed. Don't trust Clear until #1 is fixed — close the tab.

[...]

PR #428, opened 2026-09-12
fix: wipe retained Vanity nodes and resume without skipping gaps

J25dunn:
## Summary

Refs #423 (F2, retained-node cleanup, and WASM artifact flow) and #425 (Vanity retained-node finding). Does not close either omnibus audit.

- Derive BIP32 nodes in caller-owned storage, then wipe the actual final child and fixed parent. Wiping a copy still left return-value copies in compiled WASM memory.
- Wipe Silent Payments scan/spend nodes before returning the public payload, including failure paths.
- Resume from the first unfinished counter, not the sum of work across workers. A stopped parallel run may repeat later-bucket work, but cannot skip a gap.
- Preserve a stopped report when another worker finishes last.
- Add compiled-memory regressions for nodes, private keys, chain codes, Silent Payments, and a full output buffer; add uneven/out-of-order worker resume cases.

## September 13 integration and CI fix

Merged current `rock` (`bc3651c`), preserving both the Vanity and merged #427 documentation. The merge and CI fix are separate commits; no force push.

The original ordinary CI failure was real: `passphrase final child node survived`. That job imported the old checked-in WASM, while the separate fresh-Rust job passed the same regression. The site build also embedded the old modules.

The workflow now builds and tests WASM once, uploads those modules, and downloads them before compiling the site. The candidate bundles the HTML, service worker, and all three matching WASM modules. Existing downstream candidate downloads therefore also supply fresh modules to source tests and the browser harness's hooks-enabled build. Post-merge publication commits these exact modules instead of compiling a second copy.

Added job-scoped guards and negative mutations for missing files in workflow declarations, missing dependencies, wrong artifact names/destinations, incorrect ordering, and accidental recompilation. All existing WASM tests remain enabled and the memory regression is unchanged.

## Validation

- `npm run build:wasm`: passed using pinned Rust 1.95.0 and workspace-local Clang 18, locked dependencies, and remapped build paths.
- The complete fresh-WASM test command used by CI: **160 passed, 0 failed**, including the retained-secret regression.
- `node --test test/validate.test.mjs`: **78 passed**.
- `npm run build`, `npm run verify`, YAML parsing, and `git diff --check`: passed.
- `npm run test:ci`: **1,307 passed, 4 skipped, 0 failed** against the rebuilt modules.
- Confirmed all three rebuilt WASM payloads are embedded byte-for-byte in the local built HTML.
- Earlier `npm run build && npm test` (before the standalone-Firefox rerun): Firefox report collection timed out locally; Chrome passed 85/86 checks, failing the previously observed enabled-Derive-Key color assertion (expected orange, got light gray; no CSS changes in this PR). The browser harness retained handles after reporting these failures and was stopped. Full browser validation is not claimed as passing.
- September 13 browser retest at `7278d1a`, using standalone Mozilla Firefox 140.14.0 ESR and the unchanged PR browser harness: `FIREFOX_BINARY=/path/to/standalone/firefox npm run test:browser` passed **86/86 Firefox and 86/86 Chrome checks**, including hosted/offline coverage; the runner exited normally. Edge was not installed. This checkout used the previously rebuilt WASM modules described above.
- GitHub CI at this commit now passes, including the Firefox integration job (run `34754824340`).
- Scope of the retest: this is an isolated browser-suite pass, not a new full-`npm test` pass. A separate local harness branch based on `rock` still reproduced the intermittent Chrome enabled-Derive-Key color assertion during its full regression run. No CSS changes, weakened assertions, or local harness changes are included in this PR.

Generated site/WASM artifacts and locale catalogs are intentionally not committed in this PR. CI supplies the rebuilt modules; local Rust changes require `npm run build:wasm` before building/testing the site. Memory checks establish the tested vectors, not guaranteed erasure of all compiler/library internals.

Clear/pagehide cancellation is merged in #427; `.elkeys` authenticity is tracked separately in #429.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,11 @@ Official website: [entropylab.online](https://entropylab.online)
 
 ## Features
 
+When a multi-worker Vanity search stops, its next counter is the first
+unfinished counter, not the sum of candidates processed across workers.
+Resuming can repeat work from later buckets, but will not skip an unfinished
+gap in the requested range.
+
 - Accepts dice rolls, coin flips, hexadecimal entropy, BIP39 seed phrases,
   extended keys, WIF keys, raw private keys, and Casascius mini private keys.
   All five BIP39 phrase lengths (12, 15, 18, 21, and 24 words) are supported
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -50,6 +50,12 @@ material. Its security posture rests on the following model:
   iOS/macOS Lockdown Mode disables WebAssembly. Exclude the site in Safari
   or use a host that can compile WASM. There is no JavaScript secp256k1
   fallback; a host that cannot run the module is treated as broken.
+- Vanity wipes its retained final child and fixed parent nodes in place;
+  matching a copy of a node and wiping that copy is not sufficient. Regression
+  tests scan WASM memory for independently derived nodes, private keys, and
+  chain codes after a grind. Passing these vectors does not guarantee erasure
+  of every compiler or crypto-library temporary; terminate the worker to
+  release its entire WASM instance.
 - Secret byte buffers are overwritten after use, on a best-effort basis. The
   WASM bindings zero every linear-memory buffer before freeing it
   (`el_free`/`psbt_free` use volatile writes) and erase their own secret
--- a/src/js/app.js
+++ b/src/js/app.js
@@ -14305,12 +14305,11 @@ function hodlRunVanity() {
       }
       if (hodlVanityStopFirst && hodlVanityRunning) hodlVanityStop();
     },
-    onDone: ({ done, stopped }) => {
+    onDone: ({ done, stopped, nextStart }) => {
       hodlVanityRunning = false;
       hodlVanityLiveRate = 0;
-      // The next run resumes where this range ended; the start field is the
-      // durable record of what has been ground.
-      let nextStart = inputs.start + done;
+      // Only a contiguous processed prefix is safe to skip. After a stop,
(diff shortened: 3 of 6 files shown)


MrHodlX:
Audited at `7278d1a` against rock `6e1f39c` (merge-base `bc3651c`, 5 behind, merges clean). Merge it.

Wipe is real. `master_node` / `ckd_priv` / `derive_path` write into caller-owned storage. `parent` is an owned `Node` wiped with `wipe_node(&mut parent)`, not moved out of an `Option`. The two surviving Copy sites (`node = parent`, and scan/spend on the SP branch) are wiped, and SP wipes before `?`.

The regression test bites. On this branch against the committed pre-fix WASM it fails with `passphrase final child node survived` — it finds the node bytes at a real offset in linear memory. Residue check, not a `wipe_node(` count. Both grind modes, the fixed parent, and SP scan/spend at recordCap 0 and 1.

F1 closed: one `npm run build:wasm`, tested `*-wasm-b64.js` uploaded, `build` downloads them, artifact job no longer recompiles. Mutation tests in `validate.test.mjs` cover the handoff.

F2 closed: resume stops at the first unfinished bucket. May repeat later-bucket work, cannot skip a gap.

`npm run test:ci`: 1311 tests, 1 fail — the new residue test against the old committed WASM, which is correct. CI compiles fresh.

Not run here: no cargo, so I did not watch the new binary pass. No browsers. Artifact handoff read from YAML, not observed.

No blockers. Does not close #423 or #425.

PR #429, opened 2026-09-12
fix: treat .elkeys imports as unverified inputs until rederived

J25dunn:
## Summary

Refs #425, finding 3 (.elkeys cached-result authenticity). This does not close the omnibus audit.

Imported files are now unverified source inputs, not restored wallets:

- Discard cached addresses, private outputs, fingerprints, file-local IDs, and verification claims. Only an allowlist of inputs/settings and display metadata is retained.
- Load an imported entry into the lab for review and fresh derivation. Only a successful derivation can become a station wallet, match an existing key, and retire the pending entry.
- Bulk add cannot install unverified inputs; ignored entries receive the same treatment.
- Preserve encryption and user-supplied inputs. Re-derivation establishes consistency, not exclusive ownership or entropy quality.

## Compatibility

New exports use input-only vault format v2 and require the updated reader. V1 imports remain supported, but their caches are discarded. Output-only caches cannot substitute for missing source inputs. Repeated imports are not deduplicated using untrusted fingerprints; matching happens only after fresh derivation.

The app package version is unchanged. UI help, README, and SECURITY document the changed workflow.

## Tests

- `npm run build`: passed.
- `npm run test:ci`: **1,275 passed, 4 skipped, 0 failed**.
- Final full run: `node --test --test-force-exit --test-concurrency=4 test/*.test.mjs`. All non-browser tests passed; the new encrypted-file import regression passed in both Firefox and Chrome. Firefox passed 87/87 browser checks; Chrome passed 86/87, failing the existing enabled-Derive-Key color assertion (also observed on #427; no CSS or that assertion is changed here). Full Node summary: 1,276 passed, 2 failed (the Chrome subtest and its parent), 5 skipped.
- Ordinary `npm test` was also attempted; its browser process retained open handles after reporting results. The final run uses Node's exit-after-completed-tests option, without skipping assertions.
- Focused parser/controller/renderer tests: 8 passed, also 8 passed against the clean merge-tree combination with #427.
- `git diff --check`: passed.

- Parser/controller regressions cover both file versions, forged flags/identities/private outputs, ignored entries, bulk add, failed derivation, and a genuine BIP84 address from the published BIP39 vector.
- Browser regression uploads a deterministically encrypted legacy file with forged cached outputs and drives the actual upload → unverified inputs → derivation workflow.
- Existing renderer escaping and encryption/backup tests remain, with cached-result expectations updated to the input-only contract.

Generated site artifacts and locale catalogs are excluded.

### DIFF
--- a/README.md
+++ b/README.md
@@ -183,13 +183,18 @@ Official website: [entropylab.online](https://entropylab.online)
   to keep it. Closing the page discards the sitting. The notebook is a
   calculator companion, not a password manager: it only stores material the
   user generated themselves.
-  The **Key manager** tab packages selected derived Key Station keys and
-  ignored-key metadata into an `.elkeys` file. It reuses the unlocked
+  The **Key manager** tab packages selected Key Station source inputs and
+  settings, including ignored entries, into an `.elkeys` file. It reuses the unlocked
   Journal's password setting, so it adds no password prompt, random salt,
-  or random nonce. Imported keys remain in Key Manager until the user chooses
-  **Use in Key Station** for one key or **Add all to Key Station** for every
-  waiting key. Adding all skips keys already in the station and leaves ignored
-  keys untouched. Deleting a Key Station tab while a Journal is open
+  or random nonce. Imports are **unverified inputs**, not ready-to-use wallets.
+  Choose **Load inputs to derive**, review the inputs/settings, then derive each
+  key in Key Station. Cached addresses, private outputs, and fingerprints from
+  legacy files are discarded; only fresh derivation can match an existing key.
+  **Add all to Key Station** applies only to already-derived keys, leaving
+  unverified and ignored entries untouched. New files use input-only format
+  version 2 (requires this updated reader); version 1 files remain importable.
+  Cached-output-only files cannot replace missing source inputs.
+  Deleting a Key Station tab while a Journal is open
   likewise removes it from the station without discarding it from Key Manager.
   The Journal also includes a paged notepad. Pages use
   the Key Station's numbered naming convention, can be added or removed with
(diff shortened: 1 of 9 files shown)


J25dunn:
September 13 local browser retest at `4d853cf`, with standalone Mozilla Firefox 140.14.0 ESR explicitly selected via `FIREFOX_BINARY` and the unchanged PR test harness. Ran `npm run test:browser` separately from the heavy unit suites.

- Firefox: **86/87 checks passed**. Base32 keyboard test failed: `getAttribute("autocapitalize")` expected `"off"`, received `"none"`.
- Chrome: **86/87 checks passed**. The previously observed enabled-Derive-Key color assertion failed: expected `rgb(255, 153, 0)`, received `rgb(235, 235, 235)`.
- Both browsers produced their reports and the runner exited normally; no forced termination. Edge was absent/skipped.

This removes the local Firefox startup/report-collection obstacle, but is **not** a green browser-suite result or a new full-`npm test` pass. Existing Firefox CI is green; these local failures remain disclosed. No source changes, assertion weakening, or harness patches were applied to this PR for the retest.

MrHodlX:
Audited at `4d853cf` against rock `6e1f39c` (merge-base `d33d1b1`, 26 behind, merges clean, also clean stacked on #428). Merge it.

`vaultEntry` is a positive allowlist. `result`, `id`, `verified`, and `importedKeyId` cannot cross. Forged `masterFingerprint` cannot alias a station key.

Forged-cache fixture behaves: v1 file with `deadbeef` / forged WIF / forged address leaves the trusted `rootXprv` alone, pending `result === null`, bulk add installs nothing. Only a successful rederive becomes a station wallet. Canonical BIP39 vector still lands on `bc1qcr8te4kr609gcawutmrza0j4xv80jy8z306fyu`. v2 export is input-only.

`npm run test:ci`: 1279 tests, 0 fail.

Not run: browsers, so the new `browser-suite.html` path is unexercised.

One low, not a blocker: repeat imports of ignored keys can grow without bound. Dedupe later.

Does not close #425. Covers finding 3 only.

PR #431, opened 2026-09-12
Fixing the layout issue that occurs after 'Derive Key' clicked

BTCMcBoatface:
Fixed issue #238 by adding min-width: 0 and overflow-x: hidden to .workspace-panel and #calc-card (and sibling tool cards) in src/css/styles.css. The wallet result rendered after key derivation contains address tables with large min-width values (660–700px) that caused the card to expand beyond the viewport on narrow screens, particularly iPhone. The parent containers had no overflow constraints, so the tables' minimum width pushed the entire layout wider than the viewport. Adding min-width: 0 allows the flex/grid layout to shrink the containers below their intrinsic content width, while overflow-x: hidden provides a containment boundary so the inner table's own overflow: auto can scroll horizontally within the card.

### DIFF
--- a/src/css/styles.css
+++ b/src/css/styles.css
@@ -1182,6 +1182,7 @@ input.bad, textarea.bad { border-color: var(--danger) !important; }
   position: relative; z-index: 1;
   border: 1px solid var(--border); border-radius: 0 0 20px 20px;
   padding: 20px 16px;
+  min-width: 0; overflow-x: hidden;
 }
 /* The tool's context sits on the panel's own ground above the card, so the
    grey surface starts at the controls rather than at the explanation. */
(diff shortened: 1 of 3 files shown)


BTCMcBoatface:
Possible side effects, though no sign of this happening during testing or any applicable elements at this time: 

- min-width: 0: Text may wrap slightly earlier in very narrow containers (e.g., long labels or monospace values), but this is generally correct behavior since the content should adapt to available space.
- overflow-x: hidden: Any absolutely positioned elements that extend past the card's horizontal edges will be clipped. Tooltips, dropdown menus, or popovers near the right edge could be affected if they overflow horizontally. However, position: fixed overlays (QR popups, expandable cell dialogs) are unaffected since fixed positioning escapes overflow clipping.

w-s-bitcoin:
Reviewed this against issue #238 and the two prior attempts (#291 and #391). The reporter explicitly confirmed that neither earlier PR resolved the page-wide expansion, while this containment approach fixed it on their deployed fork.\n\nI pushed a cleanup follow-up that:\n- removes the ineffective scattered result/secret/table width workarounds from the earlier attempts;\n- keeps #291's independently useful responsive QR sizing;\n- contains wide derived output once at the shared workspace panel instead of clipping every individual tool card; and\n- replaces the old 220px inner-result assertion with a real post-derivation 390px page-container regression test that also confirms the address table remains horizontally scrollable.\n\nNegative control: with the new workspace containment disabled, the browser test catches the original behavior (390px container expands to 616px). With containment enabled, the full required run passed: 1,290 passed, 0 failed, 2 skipped browser engines (Firefox and Edge unavailable locally); Chrome ran all 86 browser integration checks. Generated entropylab.html was restored and is not part of the diff.

PR #432, opened 2026-09-13
Fixed: Test Base32 autocapitalize through the property, not the attribute

timechainb:
The Bech32 keyboard check read getAttribute("autocapitalize") and expected "off". The HTML spec maps "off" and "none" to the same no-capitalization state, and engines store different strings for it: Chrome keeps "off", while Firefox 128 canonicalizes the attribute to "none" however it is written. The check therefore failed on Firefox 128 with the app behaving correctly.

The reflected property reports "none" in both engines, and still reports "characters" or "sentences" if the field regresses, so asserting it keeps the check meaningful without encoding engine differences.

### DIFF
--- a/test/browser-suite.html
+++ b/test/browser-suite.html
@@ -779,7 +779,12 @@
     await chooseMode("Number bases");
     document.querySelector('input[name="entropy-format"][value="base32"]').click();
     const base32 = await until(() => document.getElementById("base32"), "Base32 input");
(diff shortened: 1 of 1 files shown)


timechainb:
@w-s-bitcoin I requested your review because this fixes a test you added, not sure why I ran into the failed test first. Either way this is a simple fix.

PR #433, opened 2026-09-13
Confirm before deriving from below-recommended entropy (#416)

hotpixelgroup:
Refs #416. Re-implementation of #420 on top of the Key Station design work from #424 — the threshold predicate and test approach from #420 are carried over (per the closing review's suggestion), and both review concerns from that close are addressed.

- [x] Security-sensitive validation/parsing changes include applicable accepted and rejected test cases, or this does not apply.

## What and why

The Key Station's Derive Key button enables as soon as **any** entropy is supplied, and the per-source "below the recommended amount" hints are easy to miss. When Derive Key is pressed with less entropy than recommended:

- A modal states the estimate: *"You have provided only about X bits of entropy (recommended: Y bits for a N-word seed)."* plus the source detail (e.g. "6 of 99 recommended rolls").
- **Add More Entropy** (also Escape / backdrop click) cancels and keeps the entered rolls/cards so the user can continue.
- **I Understand, Proceed** derives the key.
- The card is a `role="dialog"` `aria-modal` overlay following the existing overlay paradigm (`address-qr.js` / `qr-references.js`), with the shared `btn secondary` / `btn primary` vocabulary and spacing tokens.

## The two points raised when #420 was closed

1. **Focus containment** — since this dialog gates key creation, Tab and Shift+Tab are trapped and cycle through the dialog's controls (checkbox → Add More Entropy → Proceed) instead of escaping to the background; initial focus lands on the safe choice and is restored on close. No focus-trap/inert vocabulary existed in the current overlays, so the shared overlay paradigm is extended here; if a design-system-level trap lands later, this module can adopt it.
2. **Dismissal persistence** — "Don't show again this session" is **in-memory only, no Web Storage**: the bypass for a security warning never outlives the page session, and the default direction is always to show the warning. This is a deliberate decision point: the issue text suggested a permanent per-user bypass — if you'd rather have permanent `localStorage` persistence (beta-banner pattern), the injectable store from #420 can be dropped back in.

## Scope decisions (unchanged from #420)

The gate applies only to the two fresh-entropy sources that can actually derive below their recommendation:

- **Hashed dice rolls** (COLDCARD base-10 and Keystone/Ian Coleman methods): warns when `rolls < config.hashRolls` (99 for 24 words, 50 for 12, …), estimate `rolls × log2(6)`.
- **Hashed card transcripts**: warns when below the recommended deal (52+6 for 24 words), without-replacement deal entropy.

Deliberately not gated (documented on the predicate): D++/BitBox dice and direct cards only enable Derive once their full construction is entered; number bases require the exact digit count; seed-phrase and private-key modes import *existing* material with no "recommended amount"; the brain wallet keeps its own explicit acknowledgement; the multisig station derives from co-signer xpubs, not user entropy. The vanity "Update key" write-back re-derives an already-created key and intentionally bypasses the modal. `validate.test.mjs`'s "intentional low-entropy recovery behavior is documented" remains accurate: low-entropy inputs are still accepted, now behind a confirmation.

## Notes

- New English strings are keyed at the call sites only; `src/locales/` is untouched — translations are left to the post-merge translation automation (missing strings fall back to English).
- Built artifact `entropylab.html` is not committed, per convention — CI rebuilds it.

## Tests

Accepted and rejected cases per the security-coverage guidance: the node suite covers the predicate's warn/no-warn boundaries for both gated sources (including 12- vs 24-word thresholds), negative cases for every non-gated source, empty inputs, session-dismissal semantics, focus-trap cycling, and the handler wiring invariant; the browser suite drives the full appear → focus-trap cycle → cancel (input kept, nothing derived) → proceed + session dismissal → bypass flow.

## Commands run

- `npm run build` — pass (`entropylab.html` 6,499,475 bytes)
- `npm test` — 1321 tests, 1315 pass, 0 fail, 6 skipped (engine skips)
- `npm run test:browser` — Chrome/Chromium: 87/87 checks passed (Firefox/Edge not installed on this machine; skipped by the harness)


J25dunn:
September 13 local browser retest at `9e24395`, with standalone Mozilla Firefox 140.14.0 ESR selected via `FIREFOX_BINARY`, running `npm run test:browser` separately from the heavy unit suites. The PR sources and browser harness were unchanged.

- Firefox: **87/87 checks passed**.
- Chrome: **86/87 checks passed**; `all six number bases preserve the same exact entropy` failed with `bin did not accept exact zero entropy`.
- Reports were collected and the runner exited normally. Edge was absent/skipped. This is not a full-suite pass, and the existing failing Firefox CI is not cleared by this local Firefox pass.

**[P2] Wait for the new bypass derivation before cleaning up** — in `test/browser-suite.html:753-755`, the second `go.click()` is followed by a wait for `#out .key-result`, but that element already exists from the first derivation. A targeted reproduction using the actual `hodlDeriveWithProgress` and `hodlStopDerivation` functions confirms that this predicate resolves while the new derivation is pending, then the following `input(dice, "")` cancels it before the derive callback runs. The test can therefore pass without proving the session-bypass derivation succeeded, and can leave asynchronous cancellation overlapping subsequent tests.

Please wait for the newly started derivation to finish (and verify its resulting output) before clearing inputs/deleting tabs. The targeted stale-result reproduction passed; the fresh browser run also reproduced the downstream zero-entropy symptom in Chrome, but that alone does not establish that this race is its only cause. No application fix or weakened assertion is proposed here.

hotpixelgroup:
Thanks for the detailed retest and the targeted stale-result reproduction — the analysis was exactly right, and the fix is now pushed in `4c3fc7a` (also rebases onto current `rock`, resolving the `test:ci` conflict with #434's `core-importdescriptors` entry by unioning both lists).

**The bypass wait now proves the new derivation finished, per your requirement.** Completion is keyed on a committed key tab — a tab only appears once a derivation completes — rather than `#out .key-result`, which as you showed is stale from the first derivation:

- The bypass step returns to the Key Station lab (`#key-tab-lab`) and derives a *different* below-recommendation transcript (`6543216`), because two discoveries fell out of instrumenting the flow: (1) deriving again from the committed key's tab updates it in place and never commits a tab (`hodlCommitDerivedKey` early-return for non-lab active keys), and (2) the dice filter strips non-1–6 digits, so my first attempt's `1234567` was sanitized back to `123456` — the same seed, which the fingerprint dedupe then correctly updated in place instead of pushing a tab. The test now asserts the transcript survives the filter verbatim, so that class of surprise fails loudly instead of timing out.
- Completion is proven three ways: a new tab commits, its fingerprint differs from the first key's, and `#out` renders the bypass key's own *"Only 7 of 99 recommended fair-die rolls"* warning — i.e. the second derivation's result, not the stale first one.
- Inputs are only cleared and tabs deleted after that proof, and cleanup now asserts only the Key Station tab remains, so nothing leaks into later tests.

Re-verified locally: `npm run test:browser` — **87/87 in Chrome/Chromium**, including our test and the previously failing downstream number-bases/vanity checks; `npm run build && npm test` — 1335 pass, 0 fail, 6 engine skips. (Firefox isn't installed on this machine, so your standalone Firefox run remains the coverage of record there.)

timechainb:
I think the meat of this PR is very good. A lot of loose ends I would want to see sorted before this merged though. This was an issue I filed so I want to help get it to the finish line. Allow me to make a commit and it should be good to go. Give me the next 24 hours...

hotpixelgroup:
@timechainb Sounds good. I'll keep my eyes peeled.

w-s-bitcoin:
Rebased onto current rock, preserving rock’s AGENTS.md and all existing tests. The CI test list retains the newer rock suites alongside the new low-entropy suite, and the current QR initializer is preserved. The new browser coverage uses stable IDs/data attributes and verifies actual derivation completion via committed tabs and fingerprints instead of pinning styling or warning copy.

Validation: npm run build && npm test passed — 1,365 tests, 1,363 passed, 0 failed, 2 skipped (Firefox and Edge absent locally). Chrome passed 89/89 integration checks. npm run verify and git diff --check passed. Generated HTML was restored and is not included in the source commits.

Per the maintainer’s direction, this is approved to merge once fresh CI passes. @timechainb (TC) will hit it with the pretty stick in a follow-up; his visual polish is still pending, not claimed complete by this merge.

AI-assisted rebase/review: GPT-5 (Codex), OpenAI.

PR #435, opened 2026-09-13
Export a watch-only Core wallet.dat from a derived multisig

MrHodlX:
Stacks on #434. After that lands this is one commit.

**What:** MS Station already exported `importdescriptors` JSON. This adds the drop-in Bitcoin Core `wallet.dat` Key Station already has — watch-only, no CLI.

**UX:** next to Copy/Save JSON:
- **Download watch-only wallet.dat**
- Filename `entropylab-msig-2of3-watch-only-wallet.dat`
- Same birthday select (genesis default)

Online node: put the file in `wallets/<name>/wallet.dat` and `loadwallet`. First `getnewaddress` must match receive index 0.

JSON stays for the Yeti 2.0 text-file path (`bitcoin-cli importdescriptors`).

**Guards**
- Never writes `walletdescriptorkey` records. `includePrivate` is ignored.
- `disable_private_keys` always.
- Capital and lowercase xprv-family refused (same family as #434).
- SLIP-132 Zpub/Ypub rewritten to xpub/tpub before storage.
- One `walletdescriptorcache` parent per co-signer (`key_exp_index` 0..n-1), receive and change.
- HD single-key records stay byte-identical to the Core 28.3 fixtures.

**Out of scope**
Private msig wallet.dat, BIP45 two-step tails, replacing the JSON export.

**Tests:** 2-of-3 fixture → 2 descriptor records, 6 cache parents matching independent CKD, no prv/mnemonic, SQLite integrity, Zpub rewrite, private key refused, HD fixtures unchanged, button on MS Station, no new workspace tab. bitcoind `loadwallet` skip unless Core is installed.

`entropylab.html` left to CI.


src/js/wallet-export.js
```diff
@@ -264,21 +306,28 @@ var hodlWalletExport = (() => {
...
+        // account-level descriptors ("…xpub/0/*"), or the descriptor root key
+        // itself when a hardened branch already moved the key to branch level
+        // ("…xpubBranch/*") — BIP32PubkeyProvider with an empty path caches
+        // its root key. Caching the wrong parent makes Core watch a different
+        // subtree than the descriptor's. Multisig writes one parent per
+        // co-signer (key_exp_index 0..n-1).
+        const branchBody = tail.branch === null ? extendedKeyBody(xpub) : deps.deriveBranchBody(xpub, tail.branch);
+        if (branchBody.length !== 74) throw new Error("wallet.dat export: branch xpub body must be 74 bytes");
+        push(
+          concat(streamString("walletdescriptorcache"), id, u32le(index)),
```
J25dunn: [P2] Preserve Core key-expression indexes for Taproot caches

This index counts only extended public keys. In the supported `tr(NUMS,sortedmulti_a(...))` descriptor, Core assigns key-expression 0 to the constant NUMS internal key, so the co-signer caches belong at 1..n, not 0..n-1. With a valid 2-of-3 fixture, Core 28.3 `getdescriptorinfo` accepts the descriptor but `loadwallet` rejects this exported database with error -4, `Wallet corrupted`. Changing only these indexes to `index + 1` for the Taproot reproduction makes the same wallet load and derive the expected address. Please account for the internal-key expression while retaining zero-based indexes for the other wrappers, and add a Taproot Core round-trip regression with exact receive/change address comparisons.

src/js/app.js
```diff
@@ -1563,9 +1575,10 @@ function hodlMsigCoreImportDescriptorsMarkup() {
...
+    <button class="btn secondary green save-wallet-dat" id="msig-download-wallet-dat" type="button">${hodlT("Download watch-only wallet.dat")}</button>
     <button class="btn secondary" id="msig-copy-importdescriptors" type="button">${hodlT("Copy Core importdescriptors")}</button>
     <button class="btn secondary green" id="msig-save-importdescriptors" type="button">${hodlT("Save Core watch-only JSON")}</button>
-    <p class="muted wallet-dat-birthday-help" id="msig-core-importdescriptors-help">${hodlT("Watch-only JSON for bitcoin-cli importdescriptors. No private keys. Online node: createwallet disable_private_keys=true blank=true, then importdescriptors. First getnewaddress must match receive index 0 here.")}</p>
+    <p class="muted wallet-dat-birthday-help" id="msig-core-importdescriptors-help">${hodlT("Watch-only Bitcoin Core wallet.dat — put it in a wallets folder and loadwallet. No private keys. JSON is for bitcoin-cli importdescriptors. First getnewaddress must match receive index 0 here.")}</p>
```
J25dunn, line 1581: [P2] Match the wallet.dat first-address guidance to its next index

The new wallet.dat action uses `branchRangeFromRows`, which sets `nextIndex` to the last displayed index plus one, so the first `getnewaddress` does not return receive index 0 as stated here. With displayed rows 0 and 1, Core 28.3 returns index 2; reproduced for native SegWit, legacy and nested SegWit and compared against Core `deriveaddresses`. If skipping displayed addresses is intentional, describe that behavior separately from the JSON import workflow instead of asking users to verify against index 0. Please assert the exact expected address in the Core test rather than only its `bcrt1q` prefix.

J25dunn (changes requested):
Reviewed `9ac3bbc` with two reproducible P2 findings below. Bitcoin Core 28.3, running on isolated regtest with fixed public test fixtures and no peer connections, rejected the Taproot export; correcting only its cache indexes in an in-memory diagnostic made it load. The non-Taproot exports loaded with private keys disabled, but their first generated address contradicts the UI guidance.

Validation: focused existing suites 48 passed / 5 Core-dependent skips; Core was then exercised separately by absolute path. Build passed. Full `npm test`: 1,329 passed, 6 skipped, 2 failed (Chrome subtest plus parent). Firefox 86/86; Chrome 85/86 with the previously intermittent enabled-Derive-Key color assertion (expected rgb(255, 153, 0), got rgb(253, 163, 28)). The runner exited normally. No PR source changes or weakened assertions were used.

MrHodlX:
Both were real. 0d5bb4f: Taproot caches sit at 1..n (NUMS keeps 0), msig nextIndex is 0 so first getnewaddress is receive 0. HD files untouched. Mind re-running Core 28.3 on the Taproot load + exact address check?

w-s-bitcoin:
Thanks for the contribution. We are keeping the v1 feature freeze focused on UI/UX improvements to existing workflows. This adds a new export capability, so we are moving it to draft and the Ooga Tracker’s Backlog Features for reconsideration after v1. This is a scope decision, not a rejection or a technical approval; the PR will remain open. Before resuming review, it will need refreshing against the then-current rock and the relevant technical checks.\n\nThe outstanding Bitcoin Core loadwallet/first-address retest requested in the existing review still needs to be completed before any future merge.

MrHodlX:
Talked to Wicked. This is in for v1.

Same export Key Station already has, for a derived msig. Watch-only only.

The Core 28.3 loadwallet / first-address retest is still the thing that has to pass before merge.

PR #436, opened 2026-09-13
Print a watch-only multisig policy sheet

MrHodlX:
Stacks on #435 (watch-only `wallet.dat`). After that lands this is one commit.

**What:** one-page watch-only policy sheet from an already-derived MS Station result. The Yeti "pass the paper around the table" step, without a CoinKite file.

**Buttons** (hidden until receive address 0 exists):
- **Print watch-only policy sheet**
- **Save watch-only policy sheet** (`entropylab-msig-2of3-watch-only-policy.txt`)

**On the page**
- EntropyLab mark
- Network — yellow/black banner if not mainnet
- `m-of-n`, script type (same UI name), sorted vs listed op
- Co-signer fingerprints + origin paths
- Descriptor checksum
- Receive address 0 + QR (existing `hodlQrSvg`)
- Path of that address
- Warning: watch-only. Cannot spend. Verify this address on every signer before funding.

Print uses existing `@media print` / `no-print`. Chrome is hidden; the sheet is the only printed node; torn down on `afterprint`. US Letter and A4.

Saved text also includes the full receive descriptor. Never a mnemonic or xprv-family key.

**Out of scope:** CoinKite `wallet.txt`, spending, BIP 388 (next PR), a new workspace tab.

`entropylab.html` left to CI.


w-s-bitcoin:
Thanks for the contribution. We are keeping the v1 feature freeze focused on UI/UX improvements to existing workflows. This adds a new export capability, so we are moving it to draft and the Ooga Tracker’s Backlog Features for reconsideration after v1. This is a scope decision, not a rejection or a technical approval; the PR will remain open. Before resuming review, it will need refreshing against the then-current rock and the relevant technical checks.\n\nThis sheet is watch-only and is not being deferred on the same secret-key-printing grounds as PR #405. Public wallet metadata can still be privacy-sensitive, and the printed address should be checked on the signers. Its stacked dependency on PR #435 also remains deferred.

PR #439, opened 2026-09-15
lightning: decode BOLT11 invoices and BOLT12 offers in the Lightning tab

MrHodlX:
Adds a deterministic, offline decoder card to the existing Lightning tab: paste a BOLT11 invoice (`lnbc`/`lntb`/`lntbs`/`lnbcrt`) or a BOLT12 offer/invoice-request/invoice (`lno`/`lnr`/`lni`) and inspect its fields. **Decode-only by design** — no invoice creation, no payment, no network; an offer's blinded paths are counted, never followed.

## What changed

- **`src/js/bolt11.js`** — original-bech32 envelope (JS polymod on public data; no 90-char limit), full tagged-field parse, and the honesty rule that an invoice is only "signature valid" once its recoverable ECDSA signature verifies — never on checksum alone. The payee node id is recovered from the signature; with a declared `n` field the signature is validated against it instead and must be low-S, per the BOLT11 reader rules. Unknown even tags hard-fail; unknown odd tags are skipped and listed. Missing payment secret, missing/duplicate description, and sub-millisatoshi amounts reject.
- **`entropylab-wasm`** — new `secp_ecdsa_recover` export (secp256k1 crate `recovery` feature) for the node-id recovery; `src/js/secp256k1.js` exposes it as `secp256k1.recover`. Artifact rebuilt with the pinned Rust 1.95.0 toolchain.
- **`src/js/bolt12.js`** — checksum-free envelope per the spec ("We currently omit the six-character trailing checksum"), `+` continuations stripped, TLV stream with bigsize types/lengths, unknown even types rejected, odd ones skipped and listed. Signatures are never verified, so BOLT12 always renders "signature not checked" — never green.
- **UI** — three explicit states (present / structurally valid / signature valid); a wrong-chain invoice (HRP vs the page network) raises a loud warning but still decodes; the payment secret is absent from the markup until an explicit reveal; Clear and the existing session wipe (pagehide/session-clear) clear the textarea, decoded state, revealed secret, and QR.
- **Tests** — the BOLT11 and BOLT12 specification vectors (accept *and* reject), a DOM-stub decode/reveal/clear/wrong-chain flow, and wiring contracts; `test:ci` carries the new suites.

## Test evidence

`npm ci` (tail):
```
npm warn install-scripts   esbuild@0.28.2 (postinstall: node install.js)
npm warn install-scripts Run `npm install-scripts ls` to review, or `npm install-scripts approve <pkg>` to allow.
```

`npm run build`:
```
Built EntropyLab v0.1.3
  entropylab.html (6513435 bytes)
  service-worker.js (1401 bytes)
```

`npm test`:
```
ℹ tests 1318
ℹ pass 1312
ℹ fail 2
✖ headless browsers run the hosted and offline suites
✖ Firefox
```

The 2 failures are **pre-existing and unrelated**: 3 Firefox integration tests (Journal notepad color value, Journal LifeHash restore, address-row QR overlay). Verified by stashing this branch's changes and re-running the suite on the clean tree — the identical 3 tests fail there too. Nothing in this branch touches those code paths.


MrHodlX:
﻿I reviewed this PR for parsing correctness, untrusted-input flow, overflow and
buffer handling, crypto-adjacency, and test coverage. Summary: **no security
defect found, and no obstacle to merging on security grounds** — the crypto is
correct, the spec conformance checks out, and there is no network egress,
entropy generation, or memory-safety problem. I found two minor robustness
issues, and opened #465 with fixes for both.

**What I verified rather than assumed**

- **Bounds handling is sound.** `readBigsize`, `pathsField` and `payinfoField`
  all re-check `offset > bytes.length` after advancing, so a hostile length
  cannot read out of bounds; the `pathsField` hop loop is bounded by a byte
  count and `payinfoField` advances at least 28 per iteration, so there is no
  infinite loop or unbounded allocation. I tried to break these and could not.
- **No integer overflow is reachable.** BOLT12 routes everything through
  `BigInt` with bigsize minimal-encoding checks; the 1023-character BOLT11 cap
  bounds tag payloads to ~1000 words, so `wordsToInt`'s `Number` arithmetic
  tops out near 3.3e10 — far below 2^53. I built validly re-signed invoices
  with widened expiry tags to force precision loss and could not.
- **The signature rules are exactly right.** With an `n` field: verify, no
  recovery, low-S enforced. Without: recover, accept high-S. Both match
  "MUST use the `n` field to validate the signature instead of performing
  public-key recovery" / "MUST perform ECDSA public-key recovery and accept
  both high-S and low-S signatures".
- **The fallback and route-hint constants are correct** (`f` versions 17/18
  per the spec; a route hint hop is 51 bytes == 51 words + 3 padding bits), and
  the required-content rules you enforce (`payment_secret` exactly once, `d`
  xor `h`, `p` multiplier last digit 0) all match the spec text.
- **No network egress, no entropy, no new deps.** Grepped both new modules for
  `fetch`/`XHR`/`WebSocket`/`EventSource`/`sendBeacon`/`URL`/dynamic `import(`
  — nothing. Only `recovery` was added to the secp256k1 features, no `rand`.
- The committed WASM is intact: decoded SHA-256 `4cb4a5d9…` matches its
  declared hash and the `secp_ecdsa_recover` export is present.

**Two minor issues (fixed in #465)**

1. A BOLT12 `offer_absolute_expiry` / `invoice_created_at` outside the range a
   `Date` can represent makes `toISOString()` throw a `RangeError`; because
   that error is unkeyed, `hodlRunLnInv` falls through to `exception.message`
   and prints a raw `Invalid time value` at the user.
2. A BOLT11 `d` description with invalid UTF-8 throws a bare `TypeError` from
   `TextDecoder` instead of a keyed failure like every other malformed field.

Neither is a security hole — both fail closed and leak nothing — but they break
the "every failure is `{ key, vars }`" contract the modules document, and they
are reachable from untrusted input. #465 keys both and adds regression tests
that fail before the change and pass after.

**The one thing I'd fix regardless:** this branch is 80 commits behind `rock`,
so the PR is conflicted on `entropylab.html`, `package.json` and
`src/shell.html`. None of my five files conflict — the decoder logic merges
cleanly — but this needs a rebase onto `rock` before it can land.

**Test coverage** is genuinely good on negative cases: the spec's invalid
vectors, junk input, and a clever checksum-preserving tag mutation to hit the
unknown-even path. The gap was exactly the two unkeyed error paths above, now
covered. The only suite failures I see are pre-existing environment ones,
identical on the base branch: `test/i18n.test.mjs` shells out to `cat` (absent
on Windows) and the headless-Firefox browser tests.

Two lower-severity observations I did **not** turn into code, because they need
a product decision: BOLT12 offers get no chain-vs-page-network warning the way
BOLT11 invoices do, and the "Structurally valid" / "Signature valid" lines are
not coupled to the signature gate in the markup (correct today only because
`bolt11Decode` throws before returning).

Co-authored-by: DeepSeek V4.1 <deepseek-noreply@deepseek.com>

Issue #93, opened 2026-08-29
Release workflows do not test the exact wallet artifacts they publish


Rob1Ham:
## Summary

The release workflow does not run its behavioral and structural gates against the exact HTML objects it publishes.

- The `build` job creates and uploads the GitHub Pages artifact.
- `test-ci`, `test-browser`, and `verify` each start from another checkout and run another build.
- Pages deploys the object from `build`, not the independently rebuilt files exercised by those jobs.
- The `artifact` job depends only on `verify`, performs yet another build, and commits that result with `[skip ci]`.

Under normal deterministic builds these copies should match, but the workflow does not establish that identity. Job-sensitive, compromised, or faulty build code can emit clean output in test jobs and different code in a publication job.

Security review status: identified through static source review at commit `853d53db21cecf6abb935a7ae578ff813a67d794` (version 0.1.3). This has not been validated with an executable proof of concept.

Related issue #58 addresses signatures and attestations for distributed HTML. This issue is distinct: signing a release object does not prove that behavioral tests exercised that same object.

## Affected Code

- `.github/workflows/ci-cd.yml:23-48` builds and uploads the Pages candidate.
- `.github/workflows/ci-cd.yml:52-86` checks out and rebuilds separately for CI and browser tests.
- `.github/workflows/ci-cd.yml:191-219` performs a third publication build, depends only on `verify`, and commits with `[skip ci]`.
- `.github/workflows/ci-cd.yml:223-260` rebuilds for structural verification, then deploys the earlier build job's Pages object.
- `scripts/verify-site.mjs:21-56` structurally checks its local rebuild but does not compare publication-object hashes.
- `scripts/build.mjs:13,47-57,71-83` executes npm-delivered esbuild code while producing the self-contained wallet.

## Security Impact

If malicious or faulty build-affecting code produces job-dependent output, the official hosted or repository-download wallet can contain code that no behavioral gate examined. For a wallet application, modified derivation, descriptor, address, or QR output can misdirect funds or create unrecoverable wallets.

This is a release-integrity weakness, not evidence that the currently published artifact is malicious or that esbuild has a vulnerability.

## Preconditions

- Build-affecting repository or locked dependency code behaves differently between jobs, or behavioral tests fail while `verify` succeeds.
- A user trusts the hosted or committed HTML without independently rebuilding and comparing it.

## Synthetic Validation Plan

1. In a test branch, make the build emit a harmless marker based on `GITHUB_JOB`.
[...]

PR #442, opened 2026-09-15
release: OpenTimestamps the tested entropylab.html

MrHodlX:
SHA256SUMS and CID prove bytes. Sigstore proves GitHub built them. An upgraded OpenTimestamps proof proves that same digest existed as of a Bitcoin block, independently of GitHub.

This is release integrity, not a calculator feature. No workspace tab. The HTML still talks to nothing.

What lands
- After the tested candidate is committed, a new `timestamp` job runs `ots stamp entropylab.html` (pinned `opentimestamps-client==0.7.2`) against the exact candidate digest (issue #93). Calendars cannot block tests, the HTML commit, or Pages.
- `entropylab.html.ots` is committed next to the sums. If a proof already covers that digest, we do not re-stamp.
- A scheduled `ots-upgrade.yml` (daily + workflow_dispatch) runs `ots upgrade` and commits only the `.ots` file. It never rebuilds HTML.
- A pending `.ots` is a calendar receipt, not a timestamp. README / llms.txt / SECURITY.md say so. Verify upgraded proofs with `ots verify` against Bitcoin Core.

Out of scope
Stamp/verify UI, calendar calls from the page, javascript-opentimestamps, waiting in CI for the next block.

First merge produces a pending proof. Upgrade completes it once a calendar has a Bitcoin attestation — usually hours, not the same CI run.


MrHodlX:
Superseded by #490, now merged as 6f22520.

This branch had drifted into a conflict with `rock` (the OTS verification paragraph in `README.md` collided with the reproducible-build section that landed afterwards), and it carried only the first of the two commits — the follow-up that keeps `ots stamp` and `ots upgrade` working past their first run was never pushed here.

#490 is these same commits replayed onto current `rock` with that conflict resolved and the follow-up included. Authorship is preserved on the original commit. Closing this one as the work is now on `rock`.

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #449, opened 2026-09-15
Pin the WASM clang toolchain so secp256k1 bytes match across machines


MrHodlX:
The WASM inside `entropylab.html` is the crypto binary. It is bitcoin-core/libsecp256k1, compiled by whatever clang is on the GitHub runner that day.

We already pin the easy half:
- Rust 1.95.0 (`entropylab-wasm/rust-toolchain.toml`)
- crates + `Cargo.lock` + `cargo build --locked`
- path remap so the builder home directory is not in the binary
- CI rebuilds WASM from source and tests that copy

We do **not** pin clang. `build-wasm` in `.github/workflows/ci-cd.yml` runs on `ubuntu-latest`. rustc uses the runner's C compiler for `secp256k1-sys`. Ubuntu moves, the WASM bytes move.

SECURITY.md already says this out loud: "Cross-machine byte identity is not claimed — the C side compiles with the builder's clang."

SHA256SUMS / Sigstore / a timestamp only prove you got the file CI uploaded. They do not prove the compiler. That is the hole.

## What "done" looks like

1. `build-wasm` runs inside a **digest-pinned** container (`image@sha256:…`), not a floating `ubuntu-latest` clang.
2. That image has one clang and one rustc. Versions written down in-tree.
3. `SOURCE_DATE_EPOCH` is set so the clock is not in the binary.
4. Two independent builds of the same commit (CI + a human on that same image) produce the **same** sha256 for:
   - `src/js/entropylab-wasm-b64.js`
   - `src/js/psbt-wasm-b64.js`
   - `src/js/vanity-wasm-b64.js`
5. Those hashes are published next to `SHA256SUMS.txt`.
6. SECURITY.md drops "cross-machine byte identity is not claimed" **only after** step 4 is true. Until then the sentence stays.

Native `npm run build:wasm` on Windows/macOS can keep using host clang for local work. Release bytes come from the pinned image only. Do not mix them.

## Out of scope

- Guix. We are not Bitcoin Core. Do not claim we are.
- Pinning Node/esbuild for the HTML wrapper is a follow-up, not this issue. The spooky part is the secp C.
- Kiosk ISO (#408).
- New calculator features.

## Why this issue exists now

So the gap is public before someone else points at the checksum and calls it a compiler pin. The code change can wait until after 0.2.0. The commitment should not.

PR #450, opened 2026-09-15
Pin the WASM clang toolchain so secp256k1 bytes match across machines (issue #449)

MrHodlX:
Implements #449. Draft while the 0.2.0 freeze is on — CI/build-only, no app code changes; merge whenever maintainers want it.

## What and why

The WASM inside `entropylab.html` is the crypto binary (bitcoin-core/libsecp256k1's C), and it was compiled by whatever clang the GitHub runner shipped that day. This stops the compiler from moving:

- **`Dockerfile.wasm`** (new): a minimal, digest-pinned build image — `ubuntu:24.04@sha256:224a18…` (manifest-list digest), apt pinned to the dated snapshot `snapshot.ubuntu.com/ubuntu/20260914T000000Z`, clang pinned to exact package versions (`clang=1:18.0-59~exp2` + `clang-18=1:18.1.3-1ubuntu1`), rustup 1.95.0 + wasm32 target (matching each crate's `rust-toolchain.toml`), Node v22.23.2. Header comments document how a maintainer updates each pin deliberately.
- **`scripts/build-wasm.mjs`**: sets `SOURCE_DATE_EPOCH` from the current commit's timestamp when unset, so the wall clock never enters the binary.
- **`ci-cd.yml` build-wasm job**: builds the pinned image, runs the WASM build inside it **twice from scratch** (`rm -rf */target` between runs) and fails the job unless both runs produce identical sha256 for all three `src/js/*-wasm-b64.js` artifacts — the in-CI proof of issue step 4. The WASM test list runs inside the image with its pinned Node. The artifact job also builds inside the pinned image, so the shipped bytes are the reproducible ones (without that, published hashes would not be human-reproducible — steps 4–5).
- **SHA256SUMS.txt**: both producers now hash the three WASM artifacts alongside `entropylab.html`.
- **SECURITY.md / README.md**: the "cross-machine byte identity is not claimed" sentence is replaced with the honest version: byte identity is now enforced for the pinned-image build; the pinned base-image digest is itself the trust anchor — a checksum after a pinned image still means you trust that image. This is not Bitcoin Core's build system and does not claim to be.

## Commands run and results

- Pin resolution verified live over HTTPS: base-image digest via the Docker Hub registry API; snapshot Release files (noble/-updates/-security) all 200; the exact clang .deb pool URL HEAD-checked 200.
- `npm run test:ci` → 1270 pass, 0 fail (1274 total, 4 skipped)
- `npm test` → the only failures are the 2 pre-existing Firefox browser-suite entries (3 assertions: journal notepad rgba, journal LifeHash restore, address-row QR), reproduced identically on the clean tree with these changes stashed — environmental, unrelated.
- `npm run build` → success; `entropylab.html` and `src/js/*-wasm-b64.js` are NOT in this diff (restored after building).
- `test/wasm-toolchain-pin.test.mjs` (new, 7 tests) guards the pins: digest-pinned FROM, dated apt snapshot, exact clang versions, Rust channel matching `rust-toolchain.toml`, the CI double-build comparison, and the SHA256SUMS contents.

## Not verified locally (first CI run is the real test)

No Docker or Rust toolchain on the authoring machine: the image build, the WASM compilation inside it, and the double-build byte comparison have not been executed locally. The pins were verified by resolving them over HTTPS. If the `clang` metapackage's dependency resolution or build time surprises, it will show in the first CI run of this PR.

Code in this PR was generated by an AI tool, per CONTRIBUTING.md's AI-only code section.


portlandhodl:
Thanks for pushing on issue #449 — the pinned-clang approach here is what forced the conversation about who actually controls the WASM bytes. Closing as superseded by #466, which just merged and covers the shared ground (build reproducibility now enforced in CI, with a runner-vs-pinned-image digest gate and a path-independence proof for the WASM builds).

That said, #466 deliberately deferred several things this PR already built, and they'd be welcome as follow-ups on the new base:

- **Dockerfile.wasm's harder pinning** — apt frozen to a dated snapshot.ubuntu.com snapshot and exact clang package versions. #466 only pins the base image digest, so apt still floats.
- **`SOURCE_DATE_EPOCH` from the commit timestamp** in build-wasm.mjs.
- **WASM artifact hashes in SHA256SUMS.txt** (both producers).
- **The artifact job building inside the pinned image**, so fresh-vs-committed WASM identity can become a hard gate (#466 ships it as a warning).

One note for a rebase if you pick any of these up: the branch is currently against a pre-candidate-flow rock (the artifact job hunk no longer applies — the candidate now carries the tested WASM modules), and the from-scratch `rm -rf */target` step will fail on the runner as written: the container runs as root, so the host-side rm can't delete root-owned target dirs. Moving the rm inside the container invocation fixes it.

Co-author credit for the overlap is on the #466 merge commit (14145fdaae79f649bc65f8317452f493c79750d1).

— Co-authored-by: Kimi <kimi-noreply@moonshot.ai>

PR #461, opened 2026-09-16
psbt-wasm: pin DoS gates (5MB / 10k pairs)

### LINKED ISSUE #355: PSBT editor: build/inspect size caps disagree (can emit a PSBT it refuses to re-inspect)
**Reported via security audit. Informational.**

The PSBT editor can emit a PSBT it then refuses to re-inspect: the build cap is 10 MB (`psbt-wasm/src/lib.rs:914`) while the inspect cap is 5 MB (`:266`), and the pair-count cap has an off-by-one (`:235` uses `>=` on inspect, `:838` uses `>` on build, so exactly 10,000 pairs is accepted on build but rejected on inspect). Fail-closed direction, boundary UX inconsistency.

**Suggested fix:** align the caps (and the off-by-one) between the two paths so build implies inspectable.

Hero-Gamer:
Follow-up to #372 by @portlandhodl, which fixed #355.

Adds the missing Rust-level regression pins for the inspector's existing DoS bounds:

* MAX_PSBT_BYTES: exactly 5 MB accepted / 5 MB + 1 rejected
* MAX_PAIRS_PER_MAP: exactly 10,000 accepted / 10,001 rejected

These complement the existing JS-level regression coverage by pinning the enforcement directly in parse_raw() and read_map().

## Test Plan

* cargo test --lib
* npm run build
* npm test

## Attribution

This PR was prepared with LLM assistance from ChatGPT and Meta AI, and both models contributed code.

Related to #355


Hero-Gamer:
Fixed 10k-pair fixture: was 2 pushes (malformed), now 3 pushes 0x01 0x00 0x00 per dummy entry like the 10,001 test. Both now correctly pin >10k not >=. 13 tests pass.

MrHodlX:
PR #461 adds Rust-level tests for existing PSBT denial-of-service protections. It does not change production behavior.

EntropyLab already limits PSBT inspection to:

Maximum file size: 5,000,000 bytes
Maximum entries in each PSBT map: 10,000 pairs

#461 adds four boundary tests directly inside the Rust PSBT module:

Exactly 5 MB is allowed past the size gate.
5 MB plus one byte is rejected.
Exactly 10,000 map entries are allowed past the count gate.
10,001 entries are rejected.

Why it matters: JavaScript tests already exercise these limits, but #461 pins the rules at the Rust function where enforcement actually occurs. That helps prevent someone from accidentally changing > to >=, moving a check, or introducing another off-by-one error.

No UI, WASM behavior, dependencies, or user-facing functionality changes. It’s test coverage only. Before merging, I’d still want one confirmed cargo test --locked --lib run because ordinary CI currently builds the Rust module but may not execute these new native Rust tests.

PR #472, opened 2026-09-18
psbt-wasm: reject v2 fields in v0 PSBTs and undefined Taproot sighash suffixes

MrHodlX:
Two spec-compliance fixes in the PSBT decoder, plus the regenerated artifact. Authored by Claude (trailers on the commits); reviewed, rebuilt, and verified here.

**1. Reject PSBT v2-exclusive fields in a v0 PSBT (BIP-370)** — a v0 PSBT carrying v2-only fields (global 0x02–0x06, input 0x0e–0x12, output 0x03–0x04) decoded and rebuilt as if valid, though BIP-370 says these "must not be included" in v0. `parse_raw()` now rejects them by name. (Previously the invalid-vector checks only fired in the other direction: v2 missing v2-required fields.)

**2. Reject a Taproot signature with an undefined sighash suffix** — `tap_sig_json()` accepted any 65th byte, so a PSBT_IN_TAP_KEY_SIG / PSBT_IN_TAP_SCRIPT_SIG with an invalid (or explicit-DEFAULT 0x00) sighash suffix decoded as a genuine signature in the editor's diagram. BIP-341 defines only {0x01, 0x02, 0x03, 0x81, 0x82, 0x83} as suffixes; 0x00 is valid only as the implicit 64-byte form. This mirrors the check the standalone inspector's JS parser already had (issue #333).

**Verification on the pinned Rust 1.95.0 toolchain:**
- `cargo test --locked`: 9/9
- The 3 new tests **fail against the old artifact and pass against the rebuilt one** (red-before-green confirmed): `psbt-wasm.test.mjs` + `psbt-viz.test.mjs` 52/52.
- `npm run reproduce`: byte-identical builds.
- `npm test`: 1368 pass, 2 fail — the 2 are the pre-existing Firefox browser-suite entries (same assertions fail on clean rock).

Code generated by AI tools (trailers on the commits).


MrHodlX:
Follow-up on the #372 review, for the record:

**Status of the original #372 blockers** — both were already fixed and merged on rock before this work, so nothing was needed there:
- Silent Payments P2TR sender-key bug: fixed in 4415537/6348fad + hardened in 4de678e (covered by test/bip352.test.mjs).
- Vacuous typed-decode gate (issue #328): fixed in a4a7b81 (covered by test/psbt-viz.test.mjs).

**The two findings this PR fixes:**
1. **BIP-370 invalid-vector acceptance.** `parse_raw()` only checked that a v0 PSBT carries an unsigned tx — it never rejected a v0 PSBT that also carries PSBT v2-exclusive fields (PSBT_GLOBAL_TX_VERSION/_FALLBACK_LOCKTIME/_INPUT_COUNT/_OUTPUT_COUNT/_TX_MODIFIABLE, or the per-input/per-output v2 fields), which BIP-370 says "must not be included" in v0. Fixed by rejecting these fields when version == 0, with global/input/output regression cases plus a clean-decode control.
2. **Residual diagram-side issue #333.** That fix only hardened the standalone inspector's hand-rolled Schnorr parser (`hodlLooksSchnorr` in src/js/psbt-schnorr.js). The PSBT editor's diagram gets its decode from the WASM crate's `tap_sig_json()`, which accepted any 65th byte on a Taproot signature — so an undefined or explicit-0x00 sighash suffix still decoded cleanly and read as "1 signature" in the editor. Fixed by applying the same sighash-byte check in `tap_sig_json()`, with a decode-level regression test and an end-to-end test (real decode → psbtVizHtml).

**Verification:** rebuilt the artifact on the pinned Rust 1.95.0 toolchain (commit 0c5c395). The three new regression tests fail against the old artifact and pass against the rebuilt one (red-before-green). Full suite: 1368 pass, 2 fail — the 2 are the pre-existing Firefox browser-suite entries, reproduced identically on clean rock.

PR #495, opened 2026-09-19
fix(online): warn on any network-served origin, not just entropylab.online

CypherBoxLLC:
## What
The permanent "do not enter secrets" banner is revealed only when `location.hostname` is `entropylab.online` (`src/js/online.js`). A user who re-hosts the self-contained `entropylab.html` on their own web server, or opens it over `http://` on a LAN, is served over the network but sees only the smaller `Online` status tag, not the prominent banner, which is exactly the case where the warning matters most.

## Change
`src/js/online.js` now keys the banner on the transport rather than on one hostname: it reveals for any non-loopback `http(s)` origin. `file://` and loopback stay local (still forceable with `?online-preview=1` for QA). The reveal unit remains storage-free and fetch-free.

## Why this fits existing behaviour
The `Online` network-status tag is already host-independent (`navigator.onLine`), and #158 established that the cautious state is what users should always see. This brings the loud banner in line with both. Complements #186, which made the banner permanent on the canonical domain.

## Tests
- `test/ui-behaviors.test.mjs` updated: a mirror host and a LAN IP now reveal the banner; `file://` and loopback stay silent; the explicit `?online-preview=1` case still works.
- `npm run build` succeeds; `npm run test:ci` passes (1385 pass, 6 skipped, 0 fail).


MrHodlX:
PR #495 fixes the big online safety warning.

Currently, the prominent:

“Online version — Do not enter seed phrases, private keys, or other wallet secrets…”

banner appears only on entropylab.online.

That means somebody could host the same EntropyLab HTML on:

another website or mirror
a private server
a LAN address such as 192.168.1.50

…and users would not receive the prominent warning, even though the page was delivered over a network.

#495 changes the rule to:

Any non-local http:// or https:// page → show the warning
Downloaded file:// version → no warning
localhost, 127.0.0.1, and ::1 → no warning
Local preview mode remains available for testing

It changes only the warning detection and its tests—no cryptography, wallet logic, networking, or layout changes. It’s a small, legitimate safety fix

PR #500, opened 2026-09-20
wipe: drop dice previousValue on pagehide, not only the visible value

### LINKED ISSUE #423: Red team: Kimi K3 full-repo pass on rock @ e225ef3

MrHodlX:
Leftover from [#423](https://github.com/OogaBoogaX/entropylab/issues/423) F3. Grids / last-words / `[data-phrase]` went in #427. `dataset.previousValue` did not.

## What

The dice input tracks edits with `dataset.previousValue`. pagehide / persisted pageshow blanked `dice.value` and left the rolls in that attribute. Clear was already fine (restore rewrites it to empty). This is the raw entropy the visible field just lost.

Same loop, same fields:

```js
if (field) {
  field.value = "";
  delete field.dataset.previousValue;
}
```

Does not close #423. The rest of that punch list is separate.

## Test

`test/secret-clear.test.mjs`: pagehide and persisted pageshow leave `dice.value === ""` and `dataset.previousValue` gone. Failed on current `rock` for the leftover, green after the delete.

`npm run build && npm test`: 1400 pass, 6 skip. The one fail here is no headless browser in this sandbox (`test/browser.test.mjs`), not this change.

cc @portlandhodl

Co-authored-by: Grok 4.20 <grok@x.ai>


MrHodlX:
ACK — reviewed by GPT-6 Astra at 49e577426d41cb12494027c2c47ac63e6d2d5b08. No blocking findings. Focused fix within scope; regression independently verified to fail before and pass after. GitHub CI passes, including browser integration. Addresses the dice previousValue leftover only; does not close #423. Ready to merge at this SHA.

PR #503, opened 2026-09-20
fix: lock password-protected journal as an in-page vault

### LINKED ISSUE #399: Red team: Kimi K3 on rock @ 813c59a
Local Kimi Code, model moonshotai/kimi-k3, branch rock, SHA 813c59a3e3abbfc4c6760cfc546a88e4315beaf4.

# EntropyLab rock red team

**SHA:** `813c59a3e3abbfc4c6760cfc546a88e4315beaf4` (rock, "Rebuild site artifact [skip ci]" — the committed `entropylab.html` reviewed is CI-built from this tree)
**Docs read:** SECURITY.md, README.md, AGENTS.md, CONTRIBUTING.md pointer, plus `git diff 9aa852ad..813c59a` over all target files (the first pass ran against the older checkout; everything below is re-verified against rock).

**Would I put a real seed in the downloaded HTML on an air-gapped machine after this pass?** Yes — with one caveat that matters operationally: the documentation claims the artifact bundles no JavaScript cryptography beyond `uqr`, and that is false — `@scure/btc-signer` is in the shipped file and decides what addresses you see and build in the PSBT editor (F1). Nothing I found leaks or misderives key material in the main flows; the failures are of the "copy overclaims what the code guarantees" kind: Lock doesn't gate what a fresh password re-opens (F2), and Update key trusts a grind-time snapshot it never re-checks (F3). On a dedicated offline machine none of these is remotely triggerable; all four need the user at the keyboard.

## Already in the model

- `navigator.onLine`-based "Offline" tag — browser signals only, disclaimed in code (`src/js/network-check.js:9-10`), fails toward Online. Admitted.
- Key Manager / journal string secrets are dereference-only after wipe (`hodlKeyManagerWipeValue`, `src/js/app.js:10328`) — inside the admitted "strings don't reliably erase" clause.
- Vanity workers hold key material until termination; stop-before-ready now terminates instead of handing over the job (`src/js/vanity.js:321-330`) — the closed #389 fix, verified present.
- PSBT prevout truth, journal offline brute-force, weak-entropy inputs, hosted cache — all as documented.

## New findings

**F1 — Medium — Artifact mismatch: docs deny the JS crypto the file ships.**
Where: `README.md:370-373` ("the only JavaScript package bundled into the artifact is `uqr`… `@noble`/`@scure` packages remain as dev-only differential test oracles, pinned and never bundled") and `README.md:391` ("The only remaining JavaScript bundled from npm is `uqr`"). Reality: `src/js/psbt-editor.js:14`, `src/js/psbt-viz.js:21`, `src/js/script-builder.js:16` import `@scure/btc-signer`; `scripts/build.mjs:85-96` bundles with no alias; the committed `entropylab.html` contains scure's `OutScript/pk: wrong key` error string.
What happens: in the PSBT editor, every displayed address is scure's decode (`psbt-editor.js:101-107`), and the "Set script" builder's address→scriptPubKey conversion is scure's (`script-builder.js:80+`). rust-bitcoin validates the PSBT *serialization* afterward but cannot catch an address decoded to the wrong hash — the semantic loop is entirely scure. A reviewer trusting the docs audits Rust for exactly the code paths that run in JS.
Why not in SECURITY.md: SECURITY.md:36-52 attributes all curve/hash/address work to the WASM crate and README says scure is never bundled; neither admits the editor exception.
Fix shape: one sentence in both docs ("the PSBT editor's script/address rendering uses the pinned, lockfiled @scure/btc-signer"), or route the editor through `addresses.js`'s WASM `addressFromScript`.

**F2 — Medium — Lock leaves the notepad live; any new password re-opens it.**
Where: `src/js/app.js:13221-13232` (`hodlJournalLock`) wipes the encrypted book, keys, and Key Manager, but never touches `hodlJournal.pages`/notepad text or `hodlJournal.stateText`; `hodlJournalWipeMem` (13284-13310) does, and Lock doesn't call it. Post-lock the cards are hidden (`hodlSyncJournalTool`, 12111-12115) and Clear journal is disabled (12966-12971). Creating a *new* journal (any password, no old one) calls `hodlJournalWipeNotebook` only (book/keys), and `hodlSyncJournalTool` then re-shows the Notepad tab with the pre-lock notes intact.
What happens: user writes seed-adjacent notes, clicks Lock, steps away; anyone (or the user, thinking it's wiped) creates a fresh journal and the previous session's notes are on screen. "Secret comes back after lock" — it never left, and the lock status copy ("Password and entries were cleared") is technically true but scoped to the book.
Why not in SECURITY.md: the model admits Lock drops Key Manager entries and that closing the page discards the journal; nowhere does it admit Lock leaves the notepad/session snapshot resident and re-exposable under a new password.
Fix shape: have Lock call the same notepad/state/log clearing WipeMem does (or run `wipeJournal(hodlJournal)`), before `hodlJournalShowWork`.

**F3 — Low — Update key applies a stale grind snapshot; the result is never re-checked against the displayed address.**
Where: `src/js/app.js:13815-13862` (`hodlVanityApplyMatch`). The source key is found by id at click time (13819) and cloned live; only `fields.pass`/`fields.account` are written from the match. The key's *other* path settings (branch, address start, purpose, coin type) are whatever the key holds *now* — and same-fingerprint edits replace a key in place, keeping its id (`hodlCommitDerivedKey`, 10390-10403). Nothing compares the re-derived wallet's address to `match.address`, while the success copy says "the Keys tab, its exports, and the Journal show this wallet."
What happens: grind a match at branch 0 / address 0; edit the source key's address start to 7 in place (fingerprint unchanged, same id); click Update key. The key re-derives to a wallet whose first address is not the displayed vanity address. Recoverable (the row shows the full path, and account-level descriptors still cover index 0), hence Low.
Why not in SECURITY.md: the model says Update key writes the match "back to the key it came from"; it doesn't admit the key's other dials may have moved.
Fix shape: snapshot the full path components in `hodlVanityRun` (they're already in `run.pathText`) and refuse — or re-derive and compare `match.address` — at apply time.

**F4 — Low — Hybrid BIP-321 URI: the on-chain address is parsed, then silently dropped.**
Where: `src/js/bip321.js:106` parses the `bitcoin:bc1q…` path; `parseRecipientLine` (116-152) never propagates it; the send UI surfaces amount, alternatives, and the Lightning-ignored note (`src/js/app.js:9416-9430`) but not the dropped address, despite the in-code invariant "a URI's payment intent is surfaced, never silently dropped."
What happens: `bitcoin:bc1qFAMILIAR?sp=sp1qATTACKER` shows only the attacker SP outputs; a sender cross-checking the familiar on-chain address isn't shown that it's unused. Mitigated: the tool builds outputs, never pays.
Fix shape: one summary line naming the ignored on-chain address, or reject hybrid URIs.

## Unproven

- **Hardened-public-key descriptor guard** (`src/js/addresses.js:143-152`): a regex pre-rejects hardened steps on public keys to avoid a rust-miniscript panic→WASM abort. I couldn't construct a false negative (x-only hex, all SLIP-132 prefixes, `*h`, multipath tails are covered; xprv is exempt because the crate derives privately). A miss is a tab-killing trap (DoS), not a wrong address. Flagged as a hedge, not a bug.
- **PSBT inspector read-only views** show stale decoded addresses after a header network switch (no `hodl:network-default` listener; only the editor re-renders, `psbt-editor.js:1216-1221`). Scripts self-describe; cosmetic, not shown to a victim as "verified". Not chased further.

## Does not close

- **F1** stays open until the docs or the bundle change; it's a doc/artifact mismatch, not a code fix.
- The SP station keeps a derived `sp1…` address + QR on screen across a mainnet→testnet picker switch (picker syncs `sp-network`, `src/js/app.js:14162-14172`, but nothing clears `#sp-out`; Keys/Multisig invalidate). The HRP self-labels, so it's an inconsistency, not a mislabel — noted for the next pass, no finding.
- `psbt-editor.js` stale-build export: verified clean (three independent gates — disabled controls, blanked textareas/QR, handler re-checks — `psbt-editor.js:647-696, 737-766`); CSS comment at `src/css/styles.css:1458` describes the old keep-dimmed behavior, stale but harmless.
- i18n catalog injection: clean — exact-set attribute allowlist, re-rendered-from-table tags, `javascript:` impossible (`src/js/i18n-sanitize.js:22-45, 147-166`); a malicious catalog can still lie in *prose*, which is translation-review territory, not markup injection.
- The `h`/`H` hardened-marker widening in `hdkey.js:226-231` is consistent with the UI parsers; no divergence found.

MrHodlX:
Replaces Journal Lock-as-wipe with Lock-as-vault. Does not close [#399](https://github.com/OogaBoogaX/entropylab/issues/399) F2 as written (that asked Lock to wipe the notepad too).

**Behavior**
- No journal password → Lock stays disabled. Passwordless journals still exist.
- With a password → Lock seals notebook + notepad + Key Manager + session snapshot/log into RAM using the existing AES-256-GCM journal keys (synthetic IV, no CSPRNG), then drops plaintext.
- Unlock in this page restores it. No file dance.
- Save file still downloads. Clear journal still wipes, including a locked vault. Closing the tab discards the ciphertext.
- Key Station keys are left on Key Station.

**Tests**
- `test/journal.test.mjs`: vault round-trip, ciphertext has no plaintext, wrong password fails, passwordless `sealVault` throws, deterministic.
- Browser suite: passwordless Lock disabled; lock/unlock restores the entry and notepad; wrong session password stays locked; existing save-file / open-file drill still passes.

cc @portlandhodl


MrHodlX:
NACK — don’t merge [PR #503](https://github.com/OogaBoogaX/entropylab/pull/503) at 7fb18df yet. Astra review found two behavioral bugs and failing tests.

Lock can lose newly accepted edits. While encryption runs, only the Lock button is disabled; the journal remains editable and pending imports remain valid. Lock then wipes those newer changes after saving an older snapshot. I reproduced an edit disappearing after Unlock. Freeze mutations and invalidate pending imports before taking the snapshot. Code
Unlock changes which keys get exported. Restoration unconditionally includes every pending key, overriding the saved selection. I reproduced one included key and one excluded key becoming both included after restoration. Preserve the selection while remapping restored IDs. Code
Five repository tests fail. Three wipe-race tests cannot load the new hodlJournalHidePlaintext helper; the secret-clear and Journal-gating assertions also need updating. The original wipe/race suites pass on the base. GitHub’s complete-suite and UI-default jobs fail too. CI

Build passes. Full local run: 1,404 passed, six skipped, six failed—five above plus the missing-browser environment failure.

Scope: this deliberately introduces a new Lock/Unlock workflow and UI. It’s broader than the earlier no-UI cleanup, and does not close #399 as written.



Not closing,. Wicked needs to review this first

MrHodlX (commented):
NACK — don’t merge this at 7fb18df. Parking it. Journal tab work stays with Wicked.

1. Lock can lose newly accepted edits. While encryption runs, only the Lock button is disabled; the journal stays editable and pending imports stay valid. Lock then wipes those newer changes after saving an older snapshot. Freeze mutations and invalidate pending imports before taking the snapshot.
2. Unlock changes which keys get exported. Restoration unconditionally includes every pending key, overriding the saved selection. Preserve the selection while remapping restored IDs.
3. Tests fail: wipe-race can’t load `hodlJournalHidePlaintext`; secret-clear and Journal-gating assertions need updating. Complete-suite and UI-default CI jobs fail too.

This is a new Lock/Unlock workflow, not #399 F2 as written. Do not close #399.

PR #504, opened 2026-09-21
Add general on-screen keyboard for non-sensitive fields (touch devices)

doggyballs:
## Problem

EntropyLab's existing on-screen keyboards deliberately scope to **sensitive** fields (seed, passphrase, private key, Base64, Bech32) with restricted charsets — and they're exactly right for that. But ~29 **non-sensitive** text/number inputs (derivation path, advanced entry indexes, vanity params, multisig thresholds) have no on-screen keyboard at all. On a touch-only kiosk — the platform this tool is designed for — those fields are visible but untouchable.

## Approach

A parallel, general keyboard that is **invisible on desktop**:

- Initializes only when `matchMedia("(pointer: coarse)")` matches — desktop users get zero DOM and zero layout change
- Reuses the existing visual language end-to-end: `.seed-keyboard` panel chrome, `.seed-keyboard-key` styling, the same toggle icon, so it renders identically to the built-in keyboards
- **Per-field layouts**: derivation path gets a purpose-built 2-row keyboard (digits; m / ' / backspace); vanity address prefix gets a lowercase bech32 layout; plain numeric fields get a digits row; everything else falls back to a full layout
- **Column-aware placement**: the panel opens below the field's row (never inside a grid cell), justified toward the field's own column — left-column fields get a left-justified keyboard, right-column fields right-justified; single-column grids place it below the grid at full width
- **Uniform keys**: fixed-width keys (58px) in every row, same size everywhere
- One shared panel, event delegation, no per-field registration; the five sensitive fields are explicitly excluded and untouched

## What this does NOT do

- Does not touch the existing restricted-charset keyboards — their security properties are a feature and stay as-is
- Adds no dependencies — pure DOM, same as the existing system

## Testing

- Unit suite: all keyboard tests pass (1323/1331 on the full suite; the 2 failures are a pre-existing headless-Chrome watchdog limitation on my build VM, identical on pristine upstream code)
- Hardware-validated on a Raspberry Pi 5 kiosk (cage + WPE WebKit via [RockOS-Pi](https://github.com/doggyballs/RockOS-Pi)): toggles, layouts, placement, and key entry verified on a real touch panel

The design discussion lives in `docs/proposal-onscreen-keyboard-all-fields.md` in the commit. Happy to reshape any of it to fit the project's conventions.


doggyballs:
Closing for now — work in progress, will reopen when ready for review.

PR #510, opened 2026-09-22
PSBT tool design pass (WIP)

timechainb:
- Split into three tabs: PSBT Inspector · PSBT Editor · Nonce Inspector. Both inspectors
    run one analysis and share one session key; the Inspector keeps a compact nonce verdict
    so a reuse warning is never lost.
  - Key input moves into an "Inspect with key (optional)" toggle and uses the Key Station
    chip picker, so any key can be chosen, not just the active one.
  - Actions stay disabled until there's something to act on; the footer holds only the
    tool's main actions.                                                               
  - Inspection report restructured: a flat security analysis with bright status colours,
    then Tx outputs / inputs / fees with counts, dividers, orange amounts and blue
    addresses.
  - BIP-85 tab reads "BIP-85 Child".

  Fixes
  - The Nonce Inspector's Jade transcript could raise the PSBT Inspector's nonce verdict
    from Incomplete to Completed.
  - "Session ended" persisted after a keyless Inspect.
  - WIF keys in chip pickers threw unhandled LifeHash errors.


timechainb:
@portlandhodl I didnt get to the editor or the nonce tab yet but I am very happy with the design pass on the PSBT Inspector tab. I wont extend this design pass to the other 2 tabs (especially editor) beofre your pod tomorrow but I thought we could get this in and at least have a prettier inspector experience in time for the publicity... lmk if you hate anything i changed, otherwise lets merge it!

w-s-bitcoin (changes requested):
I found two issues to address before merging:

1. **Cross-session nonce detection regresses in the default Inspector.** In `src/js/app.js`, the new `if (view !== "inspect")` guards around `hodlPsbtRecordNonceInspection` (PSBT and raw-transaction paths) mean that clicking Inspect in the default PSBT Inspector neither records its signatures nor compares them with previous inspections. Before this PR, every inspection did. For example, inspect file A with one ECDSA signature, then file B with the same key and r but a different digest: the default-tab workflow has no cross-session warning, and opening Nonce Inspector for B later still cannot compare against A. Please preserve the history/check for the default inspection and surface a cross-session warning there, or make the loss of that safety behavior explicit and intentional. A browser regression for the A-then-B default-tab path would help.

2. **Please remove `entropylab.html` from this PR.** It is a generated build artifact; `AGENTS.md` and `CONTRIBUTING.md` require source-only changes, and the post-merge workflow rebuilds the artifact.

The existing CI checks are green; these findings are from reviewing the changed behavior and PR file scope. AI-assisted review: GPT-5.

w-s-bitcoin (approved):
Re-reviewed after the generated HTML artifact was removed. The project owner explicitly accepts that cross-session nonce history is populated through the Nonce Inspector rather than the default PSBT Inspector, so I am withdrawing that change request. The updated source scope looks ready to merge once the required checks finish. AI-assisted review: GPT-5.

